Issue #19, opened 2024-08-04
Support for break statements inside combinational for-loops


alikates:
Hi!
While trying to synthesize this core: https://github.com/bsc-loca/core_tile, I found out that having a break statement inside a for loop is not supported. The issue I'm seeing is inside the DCache of the design, exactly [here](https://github.com/bsc-loca/cv-hpdcache/blob/4e354ed2585257af43614ab9ac856a9bb7a1451e/rtl/src/hwpf_stride/hwpf_stride_wrapper.sv#L111-L119):
```
yosys> read_slang -f filelist_fixed.f 

1. Executing SLANG frontend.
{
  "kind": "Break"
}
Source line rtl/dcache/rtl/src/hwpf_stride/hwpf_stride_wrapper.sv:116.17-116.23:                 break;
<suppressed ~141 debug messages>
ERROR: Feature unimplemented at slang_frontend.cc:1269, see AST and code line dump above
```

I' ve looked a bit at the frontend code and saw that full cases are already supported, so I think it could be possible to transform these kind of loops into a case with priority.

For now I've given a couple hours to study the code but I'm struggling a bit figuring out where to start. Let me know if you already had this in mind. Otherwise I'd appreciate a bit of guidance on where to start to better understand the code to try add support for it.

Thanks!

alikates:
Thanks! I will contact you if I have some question about the code. Anyways I think the main issue was that I wasn't really familiar with Yosys RTLIL either way, but I'm slowly learning about both!

alikates:
I guess we can close this already, right?

povik:
Yes, let's open a separate issue to track getting core_tile through yosys-slang, as that has turned out to not be a bad codebase to test the tool on.

PR #23, opened 2024-08-12
Handle break statements on for loops

alikates:
First attempt at disabling loop iterations whenever a break statement is found. In #19 it was discussed to transform the loop into a parallel case, but I ended up just adding a conditional for each iteration and adding the or-reduce with.

For now i've tested the code with:
```verilog
module forbreak (
    input logic [7:0] bits,
    output logic [2:0] encoded
);
    always_comb begin: for_with_break
        encoded = 1'b0;
        for (integer unsigned i = 0; i < 8; i++) begin
            if (bits[i]) begin
                encoded = i;
                break;
            end
        end
    end
endmodule
```
And manually checking the truth table. It would be nice to also add a test for this.

Any suggestion is welcome!

### DIFF
--- a/slang_frontend.cc
+++ b/slang_frontend.cc
@@ -820,9 +820,6 @@ public:
 		return ret;
 	}
 
-	std::vector<RTLIL::Wire*> disableLoop_ins;
-	std::vector<RTLIL::SigSpec> disableLoop_outs;
-
 	void handle_display(const ast::CallExpression &call)
 	{
 		auto cell = netlist.canvas->addCell(NEW_ID, ID($print));
@@ -982,19 +979,19 @@ public:
 		for (auto init : stmt.initializers)
 			eval(*init);
 
-		disableLoop_ins = {};
-		disableLoop_outs = {};
-
-
 		int ncycles = 0;
 
 		if (!stmt.stopExpr) {
 			diag_scope->addDiag(diag::MissingStopCondition, stmt.sourceRange.start());
 			return;
 		}
-
 		SwitchHelper b(current_case, vstate, {RTLIL::S0});
 		b.sw->statement = &stmt;
+
+		RTLIL::Wire *disable = netlist.canvas->addWire(NEW_ID_SUFFIX("disable"), 1);
+		disable->attributes[ID($nonstatic)] = 1;
+		do_simple_assign(stmt.sourceRange.start(), disable, RTLIL::S0, true);
+		eval.frames.back().loopDisableWire = disable;
 
 		while (true) {
 			RTLIL::SigSpec cv = eval(*stmt.stopExpr);
@@ -1008,28 +1005,10 @@ public:
 			if (!cv.as_const().as_bool())
 				break;
 
-			RTLIL::SigSpec disable;
-
-			if (disableLoop_outs.empty()) {
-				disable = RTLIL::S0;
-			} else {
-				disable = disableLoop_outs.back();
-			}
-
-			// add next disable wire
-			RTLIL::Wire *breakNextLoop = netlist.canvas->addWire(NEW_ID_SUFFIX("breakNextLoop"), 1);
-			RTLIL::SigSpec breakNextLoopOred = netlist.LogicOr(breakNextLoop, disable);
-			do_simple_assign(stmt.sourceRange.start(), breakNextLoop, RTLIL::S0, false);
-
-			disableLoop_ins.push_back(breakNextLoop);
-			disableLoop_outs.push_back(breakNextLoopOred);
-
-			eval.push_frame();
-			b.branch({disable}, [&]() {	// disable on previous break
+			b.branch({substitute_rvalue(disable)}, [&]() {
 				current_case->statement = &stmt.body;
 				stmt.body.visit(*this);
 			});
-			eval.pop_frame();
 
 			for (auto step : stmt.steps)
 				eval(*step);
@@ -1073,8 +1052,9 @@ public:
 
 	void handle(const ast::BreakStatement &brk)
 	{
-		log_assert(!disableLoop_ins.empty());
-		do_simple_assign(brk.sourceRange.start(), disableLoop_ins.back(), RTLIL::S1, false);
+		log_assert(!eval.frames.empty());
+		log_assert(eval.frames.back().loopDisableWire != nullptr);
+		do_simple_assign(brk.sourceRange.start(), eval.frames.back().loopDisableWire, RTLIL::S1, true);
 	}
 
 	void handle(const ast::Statement &stmt)
--- a/slang_frontend.cc
+++ b/slang_frontend.cc
@@ -991,9 +991,6 @@ public:
 		do_simple_assign(stmt.sourceRange.start(), disable, RTLIL::S0, true);
 
 		while (true) {
-			SwitchHelper b(current_case, vstate, {RTLIL::S0});
-			b.sw->statement = &stmt;
-
 			RTLIL::SigSpec cv = eval(*stmt.stopExpr);
 			if (!cv.is_fully_const()) {
 				auto& diag = diag_scope->addDiag(diag::ForLoopIndeterminate, stmt.sourceRange);
@@ -1005,18 +1002,21 @@ public:
 			if (!cv.as_const().as_bool())
 				break;
 
+			SwitchHelper b(current_case, vstate, {RTLIL::S0});
+			b.sw->statement = &stmt;
+
 			b.branch({substitute_rvalue(disable)}, [&]() {
 				current_case->statement = &stmt.body;
 				eval.push_frame(nullptr, disable);
 				stmt.body.visit(*this);
 				eval.pop_frame();
 			});
+			b.finish(netlist);
 
 			for (auto step : stmt.steps)
 				eval(*step);
 
 			ncycles++;
-			b.finish(netlist);
 		}
 		current_case = current_case->add_switch({})->add_case({});
 	}
--- a/slang_frontend.cc
+++ b/slang_frontend.cc
@@ -985,13 +985,13 @@ public:
 			diag_scope->addDiag(diag::MissingStopCondition, stmt.sourceRange.start());
 			return;
 		}
+
 		SwitchHelper b(current_case, vstate, {RTLIL::S0});
 		b.sw->statement = &stmt;
 
 		RTLIL::Wire *disable = netlist.canvas->addWire(NEW_ID_SUFFIX("disable"), 1);
 		disable->attributes[ID($nonstatic)] = 1;
 		do_simple_assign(stmt.sourceRange.start(), disable, RTLIL::S0, true);
-		eval.frames.back().loopDisableWire = disable;
 
 		while (true) {
 			RTLIL::SigSpec cv = eval(*stmt.stopExpr);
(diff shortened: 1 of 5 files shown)


slang_frontend.cc
```diff
@@ -997,14 +1003,39 @@ struct ProceduralVisitor : public ast::ASTVisitor<ProceduralVisitor, true, false
...
+				disable = RTLIL::S0;
+			} else {
+				disable = disableLoop_outs.back();
+			}
+
+			// add next disable wire
+			RTLIL::Wire *breakNextLoop = netlist.canvas->addWire(NEW_ID_SUFFIX("breakNextLoop"), 1);
+			RTLIL::Wire *breakNextLoopOred = netlist.canvas->addWire(NEW_ID_SUFFIX("breakNextLoopOred"), 1);
+			do_simple_assign(stmt.sourceRange.start(), breakNextLoop, RTLIL::S0, false);
+			netlist.canvas->addOr(NEW_ID_SUFFIX("breakNextLoopOr"), breakNextLoop, disable, breakNextLoopOred, false);
```
povik, line 1018: Please add `RTLILBuilder::LogicOr()`, and do `breakNextLoopOred = netlist.LogicOr(breakNextLoop, disable);` here.

(Right now `RTLILBuilder` is just a thin wrapper which does constant folding, which is not useful here, but in the future it may do more, so we should be using it consistently.)

slang_frontend.cc
```diff
@@ -1039,6 +1070,12 @@ struct ProceduralVisitor : public ast::ASTVisitor<ProceduralVisitor, true, false
 		stmt.symbol.visit(*this);
 	}
 
+	void handle(const ast::BreakStatement &brk)
+	{
+		require(brk, !disableLoop_ins.empty())
```
povik, line 1075: This should be a log_assert rather than require (require is for limitations of the frontend, log_assert for invariant violations)

slang_frontend.cc
```diff
@@ -997,14 +1003,39 @@ struct ProceduralVisitor : public ast::ASTVisitor<ProceduralVisitor, true, false
...
+
+			// add next disable wire
+			RTLIL::Wire *breakNextLoop = netlist.canvas->addWire(NEW_ID_SUFFIX("breakNextLoop"), 1);
+			RTLIL::Wire *breakNextLoopOred = netlist.canvas->addWire(NEW_ID_SUFFIX("breakNextLoopOred"), 1);
+			do_simple_assign(stmt.sourceRange.start(), breakNextLoop, RTLIL::S0, false);
+			netlist.canvas->addOr(NEW_ID_SUFFIX("breakNextLoopOr"), breakNextLoop, disable, breakNextLoopOred, false);
+			disableLoop_ins.push_back(breakNextLoop);
+			disableLoop_outs.push_back(breakNextLoopOred);
+
+			SwitchHelper b(current_case, vstate, disable);
```
povik, line 1022: I think we may be able to make this into a parallel case purely by: 
 * moving the SwitchHelper out of the loop
 * passing in `RTLIL::S0` instead of the disable signal into the helper's constructor
 * doing `b.sw->parallel_case = true;`
 * using `b.branch(disable, [&]() {` later on

slang_frontend.cc
```diff
@@ -820,6 +820,8 @@ struct ProceduralVisitor : public ast::ASTVisitor<ProceduralVisitor, true, false
 		return ret;
 	}
 
+	std::vector<RTLIL::Wire*> disableLoop_ins, disableLoop_outs;
```
povik, line 823: I think this will crash/miscompile if you have two nested for loops.

povik (commented):
Thanks! I left some comments

povik:
In the Croc test this change causes combinational loops for this process: https://github.com/pulp-platform/croc/blob/1b6d6daab077d80a90c17980fe3e49d090a07234/rtl/riscv-dbg/dm_mem.sv#L227

It's not immediately clear to me why.

slang_frontend.cc
alikates: Right! I'm thinking of making the push/pop_frame function save these wires in a Frame to avoid it. What do you think?

slang_frontend.cc
alikates: I was a bit afraid of using non-static signals in the case part, but this confirms to me that it's completely fine haha!

povik: Yes, I was also surprised to learn it's fine a while ago

povik: I may be mistaken in suggesting a parallel case for this. I need to look over how parallel cases are defined in RTLIL

slang_frontend.cc
povik: Saving this in a `Frame` sounds good. I suggest the `push_frame` function takes an optional "disable wire" argument.

slang_frontend.cc
povik: Yes, `parallel_case` is wrong for this, while it allows for more than one case to be active in a switch (in the sense of RTLIL CaseRule/SwitchRule), it doesn't establish priority between the assignments from different cases (which we obviously want here for assignments from later iterations to override those from preceding). We need to do one switch per iteration like you are already doing.

alikates:
I just applied some of the suggestions, but now for some reason in the proc step all the OR gates are being optimized and replace with a constant

slang_frontend.cc
povik: Having thought about this more, instead of ORing the disable signals explicitly, and using non-blocking assignments, we want to rely on the handling of blocking assignments and do the following:

 * in handling of a loop, create a single disable wire with the attribute `$nonstatic` (see `create_local` which does the same for user variables with automatic lifetime)
 * initialize the disable wire to zero by a blocking assignment to it
 * in the handling of a break statement, use a blocking statement to set the disable signal to one
 * at the start of each iteration, pass the disable wire to `substitute_rvalue` to evaluate it at that point within the procedure, and use the signal that's returned to drive the switch that disables the loop body
 
 Note that the wire itself which is created for variables with automatic lifetime isn't ever driven, it's only a placeholder to have a key into the maps inside `VariableState` to track the value that's been assigned to them (since the automatic variables cannot be referred to outside of a procedure, we need to track the value that's been assigned to them, but we don't need to create a driver for them)

povik:
> I just applied some of the suggestions, but now for some reason in the proc step all the OR gates are being optimized and replace with a constant

This is likely becase you did part of the change towards a parallel case, but see my comments in the other thread, we can't do that

slang_frontend.cc
```diff
@@ -982,20 +979,20 @@ struct ProceduralVisitor : public ast::ASTVisitor<ProceduralVisitor, true, false
...
-
 		int ncycles = 0;
 
 		if (!stmt.stopExpr) {
 			diag_scope->addDiag(diag::MissingStopCondition, stmt.sourceRange.start());
 			return;
 		}
-
 		SwitchHelper b(current_case, vstate, {RTLIL::S0});
 		b.sw->statement = &stmt;
```
povik, line 989: The SwitchHelper needs to be constructed once per iteration, we cannot reuse it

slang_frontend.cc
```diff
@@ -1008,28 +1005,10 @@ struct ProceduralVisitor : public ast::ASTVisitor<ProceduralVisitor, true, false
...
-
-			// add next disable wire
-			RTLIL::Wire *breakNextLoop = netlist.canvas->addWire(NEW_ID_SUFFIX("breakNextLoop"), 1);
-			RTLIL::SigSpec breakNextLoopOred = netlist.LogicOr(breakNextLoop, disable);
-			do_simple_assign(stmt.sourceRange.start(), breakNextLoop, RTLIL::S0, false);
-
-			disableLoop_ins.push_back(breakNextLoop);
-			disableLoop_outs.push_back(breakNextLoopOred);
-
-			eval.push_frame();
```
povik, line 1000: We lost the `push_frame()`/`pop_frame()` calls per iteration, these are important if there are variables local to the loop body

alikates: Yep, fixed!

slang_frontend.cc
alikates: Ah sorry, so then we are fully discarding the use of a switch/case?

alikates: Nvm, I get it

povik: Pretty much, or rather we need to have a new switch per iteration (like you had it originally)

slang_frontend.cc
alikates: I just added the extra parameter in push_frame

alikates:
Okay, now it looks much better. It's still showing some errors for Croc, looks like the disable signal is not being driven somehow

povik:
> But it should end up being a constant if it doesn't see any break statements, right...?

Exactly, but I just realized what this is. For the proper working of the blocking assignments/rvalue substitutions on automatic variables, we need to set `$nonstatic` to the case tree level at which the variable lives, see `create_local`. I forgot about that detail until now.

When the level is wrong, it can create spurious uses of the `$nonstatic` wire, which is what we are seeing

alikates:
Something like this? I don't fully understand what the level means? Is it the nesting level?

povik:
> Something like this? I don't fully understand what the level means. Is it the nesting level?

Yes, it's the level on the case tree constructed by the procedural visitor, it's not necessarily the same as the input nesting level (especially because we do `current_case = root_case->add_switch({})->add_case({});` occasionally)

Consider the following

```
    reg [3:0] r;
    always_ff @(posedge clk) begin
      if (...) begin
        r = ...;
      end
      foo(r);
    end
```

The signal going into the `foo` function call will be a mux between what's assigned to `r` two lines above, and the `r` wire as driven by a flip-flop. Now consider

```
    always_ff @(posedge clk) begin
      if (...) begin
        automatic reg [3:0] r;
        r = ...;
      end
      // foo(r);
    end
```

We don't want the same mux constructed here, because it will (1) create a reference to an undriven wire (`r` is an undriven wire with the `$nonstatic` attribute) and (2) it can't get referred to from `foo(r)` anyway as at that point it's out of scope. 

The level value in `$nonstatic` signifies to the `SwitchHelper` code when the variable is going out of scope on the case tree so that the mux isn't created.

slang_frontend.cc
```diff
@@ -985,7 +986,14 @@ struct ProceduralVisitor : public ast::ASTVisitor<ProceduralVisitor, true, false
 			return;
 		}
 
+		RTLIL::Wire *disable = netlist.canvas->addWire(NEW_ID_SUFFIX("disable"), 1);
+		disable->attributes[ID($nonstatic)] = current_case->level;
+		do_simple_assign(stmt.sourceRange.start(), disable, RTLIL::S0, true);
+
 		while (true) {
+			SwitchHelper b(current_case, vstate, {RTLIL::S0});
+			b.sw->statement = &stmt;
```
povik, line 995: Please move the switch helper construction and the `b.finish(netlist)` call closer to the branch call. This will create a superfluous switch at the end when we are exiting out of the loop.

alikates:
Understood, thanks!

povik (approved):
Looks good except for the one wrinkle. Do you want to write tests for this?

alikates:
Yes, I was already taking a look on how to formally prove it. I have never used Yosys' SAT solver so I'm not really sure how to do it

povik:
FWIW the test scripts will call the SAT solver for you: https://github.com/povik/yosys-slang/blob/master/tests/selects/all.tcl See how the select tests are written.

alikates:
Yeah I saw it. What I don't know is how to correctly constrain the module

slang_frontend.cc
alikates: Done!

PR #25, opened 2024-08-13
Allow bit selection on all bitstream-convertible signals

alikates:
This should allow bit and range indexing of packed structs and other similar signals.

Btw, when writing the test I noticed there seems to be an issue when lowering from `assert` with a struct member access to a `$check` cell: somehow the arguments get lost. I will open an issue on this.


povik:
See [6fa347](https://github.com/povik/yosys-slang/commit/6fa347c6bf6f8b0bc77739eb9b8b44a3142fe3db) for fix of the CI failure

alikates:
Thanks!

PR #47, opened 2024-09-07
handle simple assignments patterns on left-hand-side (lhs)

anders-code:
logic [31:0]data[4], a, b, c, d;
    assign '{ a, b, c, d } = data;

### DIFF
--- a/src/slang_frontend.cc
+++ b/src/slang_frontend.cc
@@ -1349,6 +1349,19 @@ RTLIL::SigSpec SignalEvalContext::lhs(const ast::Expression &expr)
 											expr.type->getBitstreamWidth());
 		}
 		break;
+	case ast::ExpressionKind::SimpleAssignmentPattern:
+		{
+			const ast::SimpleAssignmentPatternExpression &pattern_expr = expr.as<ast::SimpleAssignmentPatternExpression>();
+			for (auto op : pattern_expr.elements())
+				ret = {ret, lhs(*op)};
+		}
+		break;
+	case ast::ExpressionKind::Assignment:
+		{
+			const ast::AssignmentExpression &ae = expr.as<ast::AssignmentExpression>();
+			ret = lhs(ae.left());
+		}
+		break;
 	default:
 		unimplemented(expr);
 		break;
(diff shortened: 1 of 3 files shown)


anders-code:
This seems to work OK for simple cases. I'm not sure it's the right way to implement it. It appears the SimpleAssignmentPattern AST node usually contains an embedded Assignment expression, but it can contain a few other things on the LHS, including an unpacked concatenation.

What should happen is when an array is on the RHS, each array element should map to one of the top level items in the pattern expression. But since it appears that the existing logic simply flattens all the packed and unpacked dimensions out, if the LHS item is a different size than RHS, the lengths no longer match.

Is there a better way to do this? Perhaps create multiple assignments in RTLIL?

povik:
> I'm not sure it's the right way to implement it. It appears the SimpleAssignmentPattern AST node usually contains an embedded Assignment expression, but it can contain a few other things on the LHS, including an unpacked concatenation.

FWIW the constant evaluation code in slang can be used as a reference https://github.com/MikePopoloski/slang/blob/d077ed831335f9c6f3020861eb68c18ab476a44e/source/ast/expressions/AssignmentExpressions.cpp#L1628 but it needs some digging into to understand.

I can look into this more when I find time.

> Is there a better way to do this? Perhaps create multiple assignments in RTLIL?

To keep the complexity low we should keep describing the entirety of the LHS with a single `SigSpec`, possibly introducing dummy wires for padding. That should always be possible.

anders-code:
> To keep the complexity low we should keep describing the entirety of the LHS with a single `SigSpec`, possibly introducing dummy wires for padding. That should always be possible.

will look into this idea. it might be tricky to determine without visiting both halves. ugh, this area of SV is tricky.

povik:
> To keep the complexity low we should keep describing the entirety of the LHS with a single `SigSpec`, possibly introducing dummy wires for padding. That should always be possible.

OK, so this isn't always possible thanks to sign extension. Take the following:
```
logic signed [3:0] data[1];
logic signed [5:0] a;
assign '{ a } = data;
```
We cannot describe the LHS with a `SigSpec` which would match the bit length of the RHS since the top bit needs to be assigned to all of `a[5:3]`.

povik:
I've started work on this feature in 79c1616c8e20397440fc699cb16db8c855344aed. The `expr.sv` test happened to have found a bug in slang's constant evaluator: MikePopoloski/slang#1123

TODO:

- [x] fill in the testing for mismatched size (pending the slang issue is fixed)
- [x] support non-procedural LHS

povik:
https://github.com/povik/yosys-slang/pull/303 filled in the missing support

PR #71, opened 2024-12-09
Add 'install' make target

rodrigomelo9:
Hi. I recently discovered this project and added yosys-slang to one of my Dockerfiles (https://github.com/PyFPGA/containers/blob/main/recipes/synthesis). I noticed that this typical option was missing, so I added it, taking inspiration from the ghdl-yosys-plugin.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@
 SRC_DIR := $(dir $(firstword $(MAKEFILE_LIST)))
 VPATH := $(SRC_DIR)
 YOSYS_CONFIG := $(YOSYS_PREFIX)yosys-config
+PLUGINDIR:=$(shell $(YOSYS_CONFIG) --datdir)/plugins
 CXXFLAGS :=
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -53,6 +54,10 @@ clean: clean-objects
 
 clean-all: clean clean-slang
 
+install: build/slang.so
+	$(YOSYS_CONFIG) --exec mkdir -p $(DESTDIR)$(PLUGINDIR)
+	$(YOSYS_CONFIG) --exec cp $< $(DESTDIR)$(PLUGINDIR)
```
povik, line 59: We are invoking yosys-config both here and in the assignment to the PLUGINDIR variable, I think one of those is superfluous

povik (commented):
Thanks! I left one comment

Makefile
rodrigomelo9: Welcome :-) I just simplified what is done at `install`. Let me know what you think or what your proposal is.

povik (approved):
Looks fine

PR #106, opened 2025-04-03
Fix build for Windows mingw environment

mmicko:
Building plugins for Windows environment is a bit more complicated. It requires linking to additional library to enable resolving everything included in main executable. This patch adds extraction of needed link flags (location of yosys lib directory in this case) and adding just this required library. 
We can genearalize this by doing same thing as `yosys-config` also linking all from --libs but there is no need to be honest

### DIFF
--- a/cmake/FindYosys.cmake
+++ b/cmake/FindYosys.cmake
@@ -22,4 +22,20 @@ message(STATUS "yosys-config --cxxflags (filtered): ${YOSYS_CXXFLAGS}")
 add_library(yosys::yosys INTERFACE IMPORTED)
 target_compile_options(yosys::yosys INTERFACE ${YOSYS_CXXFLAGS})
 
+if(WIN32)
+    execute_process(
+        COMMAND ${YOSYS_CONFIG} --linkflags
+        OUTPUT_VARIABLE YOSYS_LINKFLAGS
+        OUTPUT_STRIP_TRAILING_WHITESPACE
+        COMMAND_ERROR_IS_FATAL ANY
+    )
+    string(REGEX REPLACE " +" ";" YOSYS_LINKFLAGS ${YOSYS_LINKFLAGS})
+    list(FILTER YOSYS_LINKFLAGS INCLUDE REGEX "^-[L]")
+    message(STATUS "yosys-config --linkflags (filtered): ${YOSYS_LINKFLAGS}")
+
+    target_link_options(yosys::yosys INTERFACE ${YOSYS_LINKFLAGS})
+
+    target_link_libraries(yosys::yosys INTERFACE yosys_exe)
+endif()
+
 set(YOSYS_DATDIR ${YOSYS_DATDIR} PARENT_SCOPE)


whitequark:
Could you show an example of the CMake output for a MinGW build? (Ideally, could you add a CI job for MinGW?)

I'm asking because I'm concerned about having unnecessary libraries (like Tcl) being included in the `yosys::yosys` dependency. I understand that `yosys_exe` is necessary, what else gets bundled in?

mmicko:
@whitequark There is currently a breakage in oss-cad-suite builds and logs are visible https://github.com/YosysHQ/oss-cad-suite-build/actions/runs/14235245313/job/39893287803

With change I have added only yosys_exe library is added, that one just have exports of all the symbols from executable enabling mingw linker to actually create a shared library, without that you get that failure as is logs with large number of symbols being unresolved. So with this we setup only library location and link this one library nothing else.

Important section from yosys Makefile creating this library is 
`EXE_LINKFLAGS := -Wl,--export-all-symbols -Wl,--out-implib,libyosys_exe.a`

whitequark:
Thanks, I understand the cause of the breakage already (I was expecting the MinGW build to have issues but I haven't had the existing knowledge to address them upfront). However, in the logs you have the broken version that doesn't call `--linkflags`, which means the logs don't answer my question. Assuming you have a local build, can you please upload the output of that?

> Important section from yosys Makefile creating this library is `EXE_LINKFLAGS := -Wl,--export-all-symbols -Wl,--out-implib,libyosys_exe.a`

This is actually exactly what I'm concerned about. These flags are appropriate for *building* `libyosys_exe.{a,dll}`, but they are not at all appropriate for *linking to* `libyosys_exe.{a,dll}`. (In fact I will argue that the design of `yosys-config` is broken because it makes this conceptual mistake for all of the flags).

Does the link fail if you only leave the additions as:

```
if (WIN32)
    target_link_libraries(yosys::yosys INTERFACE yosys_exe)
endif()
```

If so, what is the linker error?

mmicko:
Yes in the logs you can see what is going on with current master of yosys-slang, so that one does not even try to link to yosys_exe library.

If you skip taking output from ` --linkflags` you get output like :
```
[100%] Linking CXX shared library ../slang.so
/usr/lib/gcc/x86_64-w64-mingw32/13.2.1/../../../../x86_64-w64-mingw32/bin/ld: cannot find -lyosys_exe: No such file or directory
collect2: error: ld returned 1 exit status
gmake[3]: *** [src/CMakeFiles/yosys-slang.dir/build.make:221: slang.so] Error 1
gmake[2]: *** [CMakeFiles/Makefile2:219: src/CMakeFiles/yosys-slang.dir/all] Error 2
gmake[1]: Leaving directory '/work/_builds/windows-x64/yosys-slang-plugin/yosys-slang-plugin/build'
gmake[1]: *** [Makefile:136: all] Error 2
make: *** [Makefile:7: build] Error 2

```

For your convinience, when running with `--linkflags` this is what is extracted for windows.
Note that is coming from PLUGIN_LINKFLAGS and LINKFLAGS in Makefile and that is empty for all other cases
since we just use it to store library locations.

```
-- yosys-config --datdir: ${BUILD_DIR}/yosys/yosyshq/share/yosys
-- yosys-config --cxxflags (filtered): -I/work/_builds/windows-x64/yosys-slang-plugin/yosys/yosyshq/share/yosys/include;-D_YOSYS_;-I/work/_builds/windows-x64/yosys-slang-plugin/yosys/yosyshq/share/yosys/include;-DYOSYS_VER="0.51+107";-DYOSYS_MAJOR=0;-DYOSYS_MINOR=51;-DYOSYS_COMMIT=107;-D_POSIX_SOURCE;-DYOSYS_WIN32_UNIX_DIR;-DYOSYS_ENABLE_READLINE;-I/usr/x86_64-w64-mingw32/sys-root/mingw/include;-DYOSYS_ENABLE_PLUGINS;-DYOSYS_ENABLE_GLOB;-DYOSYS_ENABLE_ZLIB;-I/usr/include/tcl86;-DYOSYS_ENABLE_TCL;-DYOSYS_ENABLE_ABC;-DYOSYS_ENABLE_COVER
-- yosys-config --linkflags (filtered): -L/work/_builds/windows-x64/yosys-slang-plugin/yosys/yosyshq/lib/yosys
-- Configuring done (0.6s)
-- Generating done (0.0s)
```
For linux is empty

whitequark (approved):
LGTM, thank you for the explanations!

mmicko:
Thanks for checking :+1:

PR #107, opened 2025-04-03
Add `slang_version` pass, to display git revisions of {yosys-,}slang

### LINKED ISSUE #99: Add command for finding out yosys-slang/slang revisions
Add a Yosys command (perhaps `slang_version`) to print out the plugin version and the vendored slang version

whitequark:
Fixes #99


whitequark:
```
yosys> slang_version
yosys-slang revision 51096461a347d93294789e4fe6f68baaba67feea
slang revision f7d4b69acac6a2fbb33fbc341bb21580282850c8
```

whitequark:
> I think it would be good to shorten the hash

I'm using the long one on purpose. This avoids any ambiguity in case of future collisions. Sure, we could pick a sufficiently long prefix, but it seems pointless to shorten it in first place if we do. Other projects like Linux are moving away from abbreviating commit hashes.

> and possibly add tag information if available, other than that this is good

I checked for that and there's no tags in either slang or yosys-slang repositories, so I can't configure `git describe` to use the proper tag pattern. Once there are tags I'm happy to add that.

povik:
> I'm using the long one on purpose. This avoids any ambiguity in case of future collisions. Sure, we could pick a sufficiently long prefix, but it seems pointless to shorten it in first place if we do. Other projects like Linux are moving away from abbreviating commit hashes.

I'd slightly prefer shortening to e.g. 12 characters

> I checked for that and there's no tags in either slang or yosys-slang repositories, so I can't configure git describe to use the proper tag pattern. Once there are tags I'm happy to add that.

OK

povik (approved):
This can go in as-is or with the abbreviating if it's a simple change

whitequark:
@povik Oh, I figured out what's up with versions. Slang itself is versioned, but since you are (a) using a fork of slang and (b) did not push the tags from the slang repo into the fork, `git-describe` is not able to use the tags.

image

povik:
> @povik Oh, I figured out what's up with versions. Slang itself is versioned, but since you are (a) using a fork of slang and (b) did not push the tags from the slang repo into the fork, git-describe is not able to use the tags.

That's right, now we've finally moved to directly consuming slang upstream (0f841ee3a702cf450705ab4bf4201830c78bc2ee) so we can revisit showing tags.

PR #113, opened 2025-04-05
Treat `celldefine like (*blackbox*)

whitequark:


whitequark:
This PR is now tested to work, and is only waiting on https://github.com/MikePopoloski/slang/pull/1302 to be merged.

src/blackboxes.cc
```diff
@@ -224,6 +224,7 @@ void export_blackbox_to_rtlil(ast::Compilation &comp, const ast::InstanceSymbol
 	}
 
 	RTLIL::Module *mod = target->addModule(name);
+	mod->set_bool_attribute(ID(blackbox), true);
```
povik, line 227: This makes me think `--empty-blackboxes` did not fully work until now

povik:
I'm in the process of updating slang. Please add a minimal test for the new behavior

povik:
Slang updated

src/blackboxes.cc
povik: It's supposed to enable automatically considering empty modules for blackboxes, same as Verific or Yosys read_verilog behavior.

whitequark: I think Yosys has that behavior built-in, but it might be in the Verilog frontend rather than in the RTLIL itself

PR #114, opened 2025-04-05
Replace the use of Yosys NEW_ID with our own name generator

### LINKED ISSUE #100: Avoid usage of NEW_ID for naming
Currently the frontend uses the Yosys macro NEW_ID to generate names for some netlist elements where the original RTL source doesn't specify a name. Replace the NEW_ID usage with a more deterministic scheme which does not involve the autoidx counter and provides better stability across runs.

whitequark:
Fixes #100.


povik (approved):
Thanks, this is good

PR #122, opened 2025-04-16
Add a way to write tests for expected frontend errors

### LINKED ISSUE #86: Add tests for expected errors
Intruduce a way to test for expected errors and add a test for #50

whitequark:
Fixes #86.


povik:
Later we'll want to have a Yosys-agnostic solution for writing tests not unlike ctests in slang upstream. That means we could refer to expected `DiagCode` symbols. I don't think we can use the numeric values inside DiagCode as those aren't stable across slang versions.

PR #145, opened 2025-05-08
log_debug is not define anymore

mmicko:
This is compile fix after https://github.com/YosysHQ/yosys/pull/5078 is merged

### DIFF
--- a/src/slang_frontend.h
+++ b/src/slang_frontend.h
@@ -49,6 +49,9 @@ using Yosys::log_id;
 using Yosys::log_signal;
 using Yosys::ys_debug;
 using Yosys::ceil_log2;
+#ifndef log_debug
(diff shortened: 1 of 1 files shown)


src/slang_frontend.cc
```diff
@@ -32,6 +32,12 @@
 #include "async_pattern.h"
 #include "variables.h"
 
+#ifdef log_debug
```
povik, line 35: Thanks, would
```
#ifndef log_debug
using Yosys::log_debug
#endif
```
work too?

If yes, I'd prefer that. We can put it into `slang_frontend.h` with other `using` declarations.

mmicko: Yes, that is passing. Have updated PR

povik:
Thanks!

PR #149, opened 2025-05-15
Add support for built-in primitives

akashlevy:
Wanted to try my hand at adding a feature to `yosys-slang`! Let me know any feedback.

You can try this out and inspect the output:
```
read_slang tests/unit/primitives.sv
write_rtlil primitives.rtlil
write_verilog -norename primitives.out.v
```

### DIFF
--- a/src/slang_frontend.cc
+++ b/src/slang_frontend.cc
@@ -2680,6 +2680,112 @@ public:
 		visitDefault(sym);
 	}
 
+	void handle(const ast::PrimitiveInstanceSymbol &sym)
+	{
+		auto ports = sym.getPortConnections();
+		std::string yosys_type(sym.primitiveType.name);
+		std::string yosys_name(sym.name);
+		bool inv_y = false;
+		bool inv_en = false;
+		RTLIL::Cell *cell;
+		auto &assign = ports.front()->as<ast::AssignmentExpression>();
+		auto y = netlist.eval.connection_lhs(assign);
+		auto z = RTLIL::Const(RTLIL::Sz, 1);
+		switch (sym.primitiveType.primitiveKind) {
+		case ast::PrimitiveSymbol::PrimitiveKind::NInput:
+			{
+				// and, or, xor, xnor, nand, nor
+				// First port is output, all others are inputs
+				if (yosys_type[0] == 'n') {
+					// nand/nor === not(and/or)
+					inv_y = true;
+					yosys_type = yosys_type.substr(1);
+				}
+				if (ports.size() == 3) {
+					// word-level primitivecell for 2 inputs
+					cell = netlist.canvas->addCell("\\" + yosys_name, "$" + yosys_type);
+					cell->setPort(ID::A, netlist.eval(*ports[1]));
+					cell->setPort(ID::B, netlist.eval(*ports[2]));
+				} else {
+					// reduce_* cell for 3 or more inputs
+					cell = netlist.canvas->addCell("\\" + yosys_name, "$reduce_" + yosys_type);
+					RTLIL::SigSpec a;
+					for (auto port : ports)
+						if (port != ports.front())
+							a.append(netlist.eval(*port));
+					cell->setPort(ID::A, a);
+				}
+				cell->setPort(ID::Y, y);
+				break;
+			}
+		case ast::PrimitiveSymbol::PrimitiveKind::NOutput:
+			{
+				// buf, inv
+				// Last port is input, all others are outputs
+				cell = netlist.canvas->addCell("\\" + yosys_name, "$" + yosys_type);
+				cell->setPort(ID::A, netlist.eval(*(ports.back())));
+				cell->setPort(ID::Y, y);
+				for (auto port : ports) {
+					if (port != ports.front() && port != ports.back()) {
+						auto &assign = port->as<ast::AssignmentExpression>();
+						netlist.canvas->connect(cell->getPort(ID::Y), netlist.eval.connection_lhs(assign));
+					}
+				}
+				break;
+			}
+		default:
+			{
+				if (yosys_type == "pulldown")
+					// pulldown is equivalent to: buffer with constant 0 input
+					cell = netlist.canvas->addBuf("\\" + yosys_name, RTLIL::SigSpec(0, 1), y);
+				else if (yosys_type == "pullup")
+					// pullup is equivalent to: buffer with constant 1 input
+					cell = netlist.canvas->addBuf("\\" + yosys_name, RTLIL::SigSpec(1, 1), y);
+				else if (yosys_type.substr(0, 5) == "bufif" || yosys_type.substr(0, 5) == "notif" ||
+								 yosys_type == "pmos" || yosys_type == "rpmos" ||
+								 yosys_type == "nmos" || yosys_type == "rnmos") {
+					// These are all just tri-state buffers, with some having inverted enable/output
+					// Use $mux instead of $tribuf to avoid Yosys issues...
+					inv_y = (yosys_type.substr(0, 5) == "notif"); // notif has inverted output
+					inv_en = (yosys_type.substr(yosys_type.length() - 3) == "if0"); // *if0 has inverted en
+					inv_en |= (yosys_type.substr(yosys_type.length() - 4) == "pmos"); // pmos has inverted en
+					auto a = netlist.eval(*ports[1]);
+					auto en = netlist.eval(*ports[2]);
+					if (!inv_en)
+						cell = netlist.canvas->addMux("\\" + yosys_name, a, z, en, y);
+					else
+						cell = netlist.canvas->addMux("\\" + yosys_name, z, a, en, y);
+				} else if (yosys_type == "cmos" || yosys_type == "rcmos") {
+					// cmos (w, datain, ncontrol, pcontrol);
+					// is equivalent to:
+					// nmos (w, datain, ncontrol); pmos (w, datain, pcontrol);
+					// Use $mux instead of $tribuf to avoid Yosys issues...
+					auto a = netlist.eval(*ports[1]);
+					auto n_en = netlist.eval(*ports[2]);
+					auto p_en = netlist.eval(*ports[3]);
+					auto nmos = netlist.canvas->addMux("$" + yosys_name + "_nmos", a, z, n_en, y);
+					auto pmos = netlist.canvas->addMux("$" + yosys_name + "_pmos", z, a, p_en, y);
+					transfer_attrs(sym, nmos);
+					cell = pmos; // transfer_attrs to pmos after switch block
+				}
+				else {
+					// bidir (tran/rtran/tranif0/rtranif0/tranif1/rtranif1) are unsupported
+					unimplemented(sym); 
+				}
+			}
+		}
+		cell->fixup_parameters();
+		transfer_attrs(sym, cell);
+		if (inv_y) {
+			// Invert output signal where needed
+			netlist.canvas->rename(cell->name, "$" + yosys_name + "_yinv");
+			auto mid_wire = netlist.canvas->addWire("$" + yosys_name + "_mid", y.size());
+			auto inv_cell = netlist.canvas->addNot("\\" + yosys_name, mid_wire, y);
+			cell->setPort(ID::Y, mid_wire);
+			transfer_attrs(sym, inv_cell);
+		}
+	}
+
 	void handle(const ast::Symbol &sym)
 	{
 		unimplemented(sym);
--- a/src/slang_frontend.cc
+++ b/src/slang_frontend.cc
@@ -2706,7 +2706,7 @@ public:
 					yosys_type = yosys_type.substr(1);
 				}
 				if (ports.size() == 3) {
-					// word-level primitivecell for 2 inputs
+					// word-level primitive cell for 2 inputs
 					cell = netlist.canvas->addCell("\\" + yosys_name, "$" + yosys_type);
 					cell->setPort(ID::A, netlist.eval(*ports[1]));
 					cell->setPort(ID::B, netlist.eval(*ports[2]));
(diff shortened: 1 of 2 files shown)


src/slang_frontend.cc
```diff
@@ -2684,6 +2684,123 @@ struct PopulateNetlist : public TimingPatternInterpretor, public ast::ASTVisitor
...
+					transfer_attrs(sym, nmos);
+					cell = pmos; // transfer_attrs to pmos after switch block
+				}
+				else {
+					// bidir (tran/rtran/tranif0/rtranif0/tranif1/rtranif1) are unsupported
+					unimplemented(sym); 
+				}
+			}
+		}
+		if (cell->type != ID($buf)) // backwards compatibility for Yosys versions < 0.46
```
povik, line 2791: We can drop support for < 0.46 if it's too cumbersome to support. Another solution is we can use `$pos` or single-bit `$_BUF_` or plain connections here.

akashlevy: No I got it, I think :)

akashlevy: Oh $pos is brilliant... maybe we should use that

akashlevy: Looks like it's going thru now

src/slang_frontend.cc
```diff
@@ -2684,6 +2685,120 @@ struct PopulateNetlist : public TimingPatternInterpretor, public ast::ASTVisitor
...
 
+	void handle(const ast::PrimitiveInstanceSymbol &sym)
+	{
+		auto ports = sym.getPortConnections();
+		std::string yosys_type(sym.primitiveType.name);
+		std::string yosys_name(sym.name);
+		bool inv_y = false;
+		bool inv_en = false;
+		RTLIL::Cell *cell;
+		auto &assign = ports.front()->as<ast::AssignmentExpression>();
```
povik, line 2696: It'd be best to put `ast_invariant(sym, ports.front().kind == ast::ExpressionKind::Assignment);` above this to document the assumption

src/slang_frontend.cc
```diff
@@ -2684,6 +2685,120 @@ struct PopulateNetlist : public TimingPatternInterpretor, public ast::ASTVisitor
...
+		auto y = netlist.eval.connection_lhs(assign);
+		auto z = RTLIL::Const(RTLIL::Sz, 1);
+		switch (sym.primitiveType.primitiveKind) {
+		case ast::PrimitiveSymbol::PrimitiveKind::NInput:
+			{
+				// and, or, xor, xnor, nand, nor
```
povik, line 2702: I think I'd prefer explicit matching on `type.name` over splicing and analyzing the strings. Something like:
```
			RTLIL::IdString op;
			bool negate = false;
			if (!type.name.compare("and")) {
				op = ID($reduce_and);
			} else if (!type.name.compare("or")) {
				op = ID($reduce_or);
			} else if (!type.name.compare("nand")) {
				op = ID($reduce_and);
				negate = 1;
			} else if (!type.name.compare("nor")) {
				op = ID($reduce_or);
				negate = 1;
			} else if (!type.name.compare("xor")) {
				op = ID($reduce_xor);
			} else if (!type.name.compare("xnor")) {
				op = ID($reduce_xnor);
			} else {
				ast_unreachable(sym);
			}
```
(I'm pasting this from an old unfinished patch which was supposed to add primitives support.)

src/slang_frontend.cc
```diff
@@ -2684,6 +2685,120 @@ struct PopulateNetlist : public TimingPatternInterpretor, public ast::ASTVisitor
...
+				// First port is output, all others are inputs
+				if (yosys_type[0] == 'n') {
+					// nand/nor === not(and/or)
+					inv_y = true;
+					yosys_type = yosys_type.substr(1);
+				}
+				if (ports.size() == 3) {
+					// word-level primitive cell for 2 inputs
+					cell = netlist.canvas->addCell("\\" + yosys_name, "$" + yosys_type);
```
povik, line 2711: Use `netlist.id(sym)` to get the name for the cell -- it includes the hierarchical path as appropriate

src/slang_frontend.cc
```diff
@@ -2684,6 +2685,120 @@ struct PopulateNetlist : public TimingPatternInterpretor, public ast::ASTVisitor
...
+					// pullup is equivalent to: buffer with constant 1 input
+					cell = netlist.canvas->addCell("\\" + yosys_name, ID($buf));
+					cell->setPort(ID::A, RTLIL::SigSpec(1, 1));
+					cell->setPort(ID::Y, y);
+				}
+				else if (yosys_type.substr(0, 5) == "bufif" || yosys_type.substr(0, 5) == "notif" ||
+								 yosys_type == "pmos" || yosys_type == "rpmos" ||
+								 yosys_type == "nmos" || yosys_type == "rnmos") {
+					// These are all just tri-state buffers, with some having inverted enable/output
+					// Use $mux instead of $tribuf to avoid Yosys issues...
```
povik, line 2759: Btw what are those Yosys issues? I know the tri-state support in Yosys is limited but I've mostly been avoiding it.

src/slang_frontend.cc
```diff
@@ -2684,6 +2685,120 @@ struct PopulateNetlist : public TimingPatternInterpretor, public ast::ASTVisitor
...
+					auto n_en = netlist.eval(*ports[2]);
+					auto p_en = netlist.eval(*ports[3]);
+					auto nmos = netlist.canvas->addMux("$" + yosys_name + "_nmos", z, a, n_en, y);
+					auto pmos = netlist.canvas->addMux("$" + yosys_name + "_pmos", a, z, p_en, y);
+					transfer_attrs(sym, nmos);
+					cell = pmos; // transfer_attrs to pmos after switch block
+				}
+				else {
+					// bidir (tran/rtran/tranif0/rtranif0/tranif1/rtranif1) are unsupported
+					unimplemented(sym);
```
povik, line 2784: I want to avoid new `unimplemented` calls

If useful I have these diagnostics on file in the old patch:

```
// definition
	slang::DiagCode UdpUnsupported(slang::DiagSubsystem::Netlist, 1043);
	slang::DiagCode PrimTypeUnsupported(slang::DiagSubsystem::Netlist, 1044);

// setup_messages
		engine.setMessage(UdpUnsupported, "user-defined primitives unsupported");
		engine.setSeverity(UdpUnsupported, slang::DiagnosticSeverity::Error);

		engine.setMessage(PrimTypeUnsupported, "primitives of type '{}' unsupported");
		engine.setSeverity(PrimTypeUnsupported, slang::DiagnosticSeverity::Error);
```
The diag code number will probably need adjusting.

src/slang_frontend.cc
```diff
@@ -2684,6 +2685,120 @@ struct PopulateNetlist : public TimingPatternInterpretor, public ast::ASTVisitor
...
+					cell = pmos; // transfer_attrs to pmos after switch block
+				}
+				else {
+					// bidir (tran/rtran/tranif0/rtranif0/tranif1/rtranif1) are unsupported
+					unimplemented(sym); 
+				}
+			}
+		}
+		if (cell->type == ID($buf) && !Yosys::yosys_celltypes.cell_known(cell->type))
+			cell->type = ID($pos); // backwards compatibility for yosys < 0.46 (no $buf cells)
```
povik, line 2789: Good call on querying the cell types but using `$buf` otherwise

tests/unit/primitives.sv
```diff
@@ -0,0 +1,213 @@
...
+  // // rtranif1: bidirectional tribuf
+  // inout logic rtranif1_a,
+  // input logic rtranif1_en,
+  // inout logic rtranif1_y0
+);
+
+  // and
+  and and_inst0(and_y0, and_a);
+  and and_inst1(and_y1, and_a, and_b);
+  and and_inst2(and_y2, and_a, and_b, and_c);
```
povik, line 152: If you want you can add functional assertions here and those should be tested automatically, e.g.
```
// don't assert anything in presence of x-bits
always_comb if (^{and_a, and_b, and_c} !== 'x) begin
  assert(and_y0 == and_a);
  assert(and_y1 == and_a && and_b);
  assert(and_y2 == and_a && and_b && and_c);
end
```

src/slang_frontend.cc
```diff
@@ -2684,6 +2685,120 @@ struct PopulateNetlist : public TimingPatternInterpretor, public ast::ASTVisitor
...
+	{
+		auto ports = sym.getPortConnections();
+		std::string yosys_type(sym.primitiveType.name);
+		std::string yosys_name(sym.name);
+		bool inv_y = false;
+		bool inv_en = false;
+		RTLIL::Cell *cell;
+		auto &assign = ports.front()->as<ast::AssignmentExpression>();
+		auto y = netlist.eval.connection_lhs(assign);
+		auto z = RTLIL::Const(RTLIL::Sz, 1);
```
povik, line 2698: We don't need this, it should be possible to use `RTLIL::Sz` directly for `addMux` argument

povik (commented):
Thanks Akash, leaving feedback

tests/unit/primitives.sv
akashlevy: That's a cool feature!

src/slang_frontend.cc
akashlevy: There's a bug (that still exists) where if `fixup_parameters()` is called on a `$tribuf` it adds parameters that don't exist like `A_WIDTH`, `Y_WIDTH`...

src/slang_frontend.cc
akashlevy: Sure, can do. I was just trying to trim the code a bit

src/slang_frontend.cc
povik: If it's just about `fixup_parameters()` not working then I think we should use `$tribuf` and avoid the fixup call on the cell. It's likely the mux against a z-value would not get handled as expected in a flow.

src/slang_frontend.cc
akashlevy: Thanks, still learning the best conventions here :)

src/slang_frontend.cc
akashlevy: I think it would probably get handled better than tribufs do. For example, constant propagation is not implemented for tribuf enables, but it is for muxes.

Two other points in favor of muxes:
1. It is possible to implement to implement these tribufs with a single cell, which avoids adding a separate inverter for the enable signal
2. If `$tribuf` cells are needed, the user could run the `tribuf` Yosys pass, which should convert `$mux` cells with a `'z` input to `$tribuf`

akashlevy: BTW I initially implemented it with tribufs before I realized how truly poor the Yosys support is

tests/unit/primitives.sv
akashlevy: Added them :)

src/slang_frontend.cc
akashlevy: No more string slicing anywhere in this patch :)

src/slang_frontend.cc
akashlevy: Added and using `netlist.add_diag` now

src/slang_frontend.cc
akashlevy: Thanks, this is pretty elegant imo

src/slang_frontend.cc
akashlevy: Thanks for the pointer, fixed

akashlevy:
Ready for review again

src/slang_frontend.cc
povik: I understand your points but I believe the `$tribuf` inference needs to happen inside the frontend and cannot be inferred from RTLIL as an afterthought because depending on the ordering of passes there's high risk of miscompilation.

Most Yosys passes cannot be trusted to handle the duplicate `$mux` drivers on a tristate line correctly. It helps if you can call `tribuf` just after the frontend, but with yosys-slang we don't give users this option because we call a bunch of passes internally, among them `opt_expr -keepdc`.

I've found e.g. the following:

```
module top(output y, input x, input f, output m);
	wire en1 = 0;
	assign y = en1 ? x : 'z;
	wire en2 = 1;
	assign y = en2 ? f : 'z;

	assign m = f;
endmodule
```

Due to `opt_clean` and `opt_expr` behavior, if you do `read_slang input.sv; opt_clean;` the passthrough from `f` to `m` gets severed and instead there's `assign f = 'z;`.

I think it's better for users to get an unmapped `$tribuf` at the end of their flow than to risk silent miscompilation. I've opened #152 to deal with tri-state drivers generally.

povik:
> Probably should have asked if you'd started a patch

No worries. Who knows when I'd finish it!

src/slang_frontend.cc
akashlevy: What if we called `tribuf` as one of the passes in the frontend?

povik: That's a good suggestion, I'll think it over

akashlevy: One last thing to note is that the issue with `assign m = 'z;` is a Yosys bug, as `'z` should be assigned with lowest priority always. The tri-state support is a problem beyond just the way `$tribuf` cells are handled.

povik: It might make sense for `'z` to be assigned with lowest priority but I don't think that's specified semantics in RTLIL.

To move this PR forward we can insert `tribuf` into the internal read_slang script (below `proc_clean`) and comment out parts of the unit test which cause the SAT crash for now.

akashlevy: Here's the SAT error btw:
```
ERROR: Failed to import cell $318 (type $tribuf) to SAT database.
```

akashlevy: Done!

PR #190, opened 2025-06-12
tests: run in cmake for consistency

donn:
- add ys/tcl script tests to cmake -- now using cmake variables to find built shared object and yosys
- cmake/FindYosys.cmake: add and elevate YOSYS_BINDIR
- updated CI to use `ctest` instead of custom test command

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -1,8 +1,35 @@
-file(GLOB YS_TESTS "${CMAKE_CURRENT_SOURCE_DIR}/*/*.ys")
-file(GLOB TCL_TESTS "${CMAKE_CURRENT_SOURCE_DIR}/*/*.tcl")
-
-set(ALL_TESTS ${YS_TESTS} ${TCL_TESTS})
+set(ALL_TESTS
+    unit/async.ys
+    unit/function_call.ys
+    unit/latch.ys
+    unit/selftests.tcl
+    various/blackbox_scenarios.ys
+    various/defaults.ys
+    various/delays.ys
+    various/expr.ys
+    various/flop_naming.ys
+    various/formal_stmts.ys
+    various/intf_array_naming.ys
+    various/intf_w_hierarchy.ys
+    various/issue142.ys
+    various/issue50.ys
+    various/mem_inference.ys
+    various/meminit.ys
+    various/pragmas.ys
+    various/regress.ys
+    various/stringattrs.ys
+    various/stringparams.ys
+    various/timescale.ys
+    various/top_attr.ys
+    various/unknown_cells.ys
+)
 foreach(test_script ${ALL_TESTS})
-    get_filename_component(test_dir ${test_script} DIRECTORY)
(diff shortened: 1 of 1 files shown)


donn:
All green. https://github.com/donn/yosys-slang/actions/runs/15606199359

povik:
> Sorry for the CI failures, @povik, was still debugging things. I'll take care in the future to not open a PR until I've made sure the CI has run successfully in my repo.

Not at all, feel free to use the CI on draft PRs as much as you need. I'll let @whitequark look the change over as she's familiar with cmake and the best practices.

.github/workflows/build-and-test.yaml
```diff
@@ -140,8 +140,10 @@ jobs:
           cd tests/third_party/yosys
           sudo make install
       - name: Run tests
+        shell: bash
         run: |
-          tests/run.sh
+          cd build
+          ctest --rerun-failed --output-on-failure
```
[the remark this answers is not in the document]
povik: Let's defer until we merge and I give ctest a spin

tests/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+file(GLOB YS_TESTS "${CMAKE_CURRENT_SOURCE_DIR}/*/*.ys")
+file(GLOB TCL_TESTS "${CMAKE_CURRENT_SOURCE_DIR}/*/*.tcl")
```
whitequark, line 2: Avoid use of globs in CMake. If you use a glob, when you add a new test and re-run CMake, it won't see it and everything will appear green even though the code suggests it should be red. For the same reason I haven't used globs for the sources.

tests/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
...
+
+set(ALL_TESTS ${YS_TESTS} ${TCL_TESTS})
+foreach(test_script ${ALL_TESTS})
+    get_filename_component(test_dir ${test_script} DIRECTORY)
+    add_test(NAME ${test_script} WORKING_DIRECTORY ${test_dir} COMMAND  stdbuf -oL -eL ${YOSYS_BINDIR}/yosys -m $<TARGET_FILE:yosys-slang> ${test_script})
```
whitequark, line 7: - What's `stdbuf` for? I'm not sure how portable yosys-slang is currently, but unconditional use of a coreutils-only utility would almost certianly make it less portable than before.
- Please split this long invocation into multiple lines, like `execute_process` above.

tests/CMakeLists.txt
povik: Is there a way to make it pick up all tests without listing them?

whitequark: No, listing them is the whole point.

tests/CMakeLists.txt
donn: https://github.com/povik/yosys-slang/issues/192

whitequark (approved):

PR #202, opened 2025-06-30
Add code coverage collection using codecov.io

### LINKED ISSUE #137: Set up codecov.io
Set up collection of test coverage inside CI and upload to codecov.io where the results can be browsed

whitequark:
This PR is using the token-less method of coverage collection, to include coverage data from PR branches in forks.

Fixes #137.

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -67,6 +67,7 @@ jobs:
       - name: Set up ccache
         run: |
           ccache --max-size=2G -z
+      # This cache is shared with the `coverage.yaml` workflow.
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-test.yaml
```diff
@@ -88,7 +89,7 @@ jobs:
           cd tests/third_party/yosys
           make config-gcc
           echo "ENABLE_ABC := 0" >> Makefile.conf
-          make -j6
+          make -j$(nproc)
```
povik, line 92: I think I overrode nprocs because of OOMs. Maybe conditions changed and we don't need the override, we'll see

PR #220, opened 2025-08-21
Replace CMAKE_SOURCE_DIR with CMAKE_CURRENT_SOURCE_DIR

ItchyTrack:
Replacing CMAKE_SOURCE_DIR will allow it to be using as a library in other CMake projects

This is because when yosys-slang is included/added as a subdirectory CMAKE_SOURCE_DIR will point to the wrong directory (The top CMakeLists.txt) causing the config to fail.

CMAKE_CURRENT_SOURCE_DIR will point to the directory that the current CMake is in and so should not change functionality when building regularly.

Edit:
Also swapped ${CMAKE_SOURCE_DIR}/cmake" and ${CMAKE_MODULE_PATH} because this allows other libs to override where find_package looks by setting CMAKE_MODULE_PATH.

This should pr should probably have a different name


povik:
Thanks, I'll let @whitequark take a look

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-if (CMAKE_BINARY_DIR STREQUAL CMAKE_SOURCE_DIR)
+if (CMAKE_BINARY_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
```
whitequark, line 1: This doesn't look right to me.

CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
 project(yosys-slang CXX)
-set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake" ${CMAKE_MODULE_PATH})
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
```
whitequark, line 14: I don't think we *want* the modules to be overridden; this makes accidental name collisions possible where previously they were not

whitequark (changes requested):

CMakeLists.txt
ItchyTrack: This says that if the CMAKE_BINARY_DIR is the directory of yosys-slang then you cant build. This makes sense because you could have
./top/CMakeLists.txt                        # user cmake
./top/yosys-slang/CMakeLists.txt  # yosys-slang cmake

when in top/yosys-slang/CMakeLists.txt
CMAKE_SOURCE_DIR is set to "./top"
CMAKE_CURRENT_SOURCE_DIR is set to "./top/yosys-slang"

if they set CMAKE_BINARY_DIR to "./top/yosys-slang" and we are comparing CMAKE_SOURCE_DIR then it would allow the build to be "In-tree". Which is not supported according to the fatal error.

Comparing CMAKE_CURRENT_SOURCE_DIR instead the makes the STREQUAL true and calls the FATAL_ERROR stopping the build from being "In-tree".

When CMAKE_SOURCE_DIR is equal to CMAKE_CURRENT_SOURCE_DIR this will have the same behavior as before the change.

Issue #84, opened 2025-01-23
support `$past` and related system functions


rroohhh:

povik:
Is $past an SVA feature, or can you use it in a procedural context everywhere?

rroohhh:
1800-2023 says at `16.9.3 Sampled value functions`:
> The use of these functions is not limited to assertion features; they may be used as expressions in procedural code as well.

concerning these system functions:
```systemverilog
$sampled ( expression ) 
$rose    ( expression [ , [ clocking_event ] ] ) 
$fell    ( expression [ , [ clocking_event ] ] ) 
$stable  ( expression [ , [ clocking_event ] ] ) 
$changed ( expression [ , [ clocking_event ] ] ) 
$past    ( expression1 [ , [ number_of_ticks ] [ , [ expression2 ] [ , [ clocking_event ] ] ] ] ) 
```

CheeksTheGeek:
Given $past is probably the main one to work on here in terms of thinking the design for it (since all other depend on it according to LRM), some questions:
- would it be okay if we put a max depth for the $past()? like I don't think there would ever be a need of $past(x, 1000). And so, would a good maximum be 256? Or is that a bad practice for shift register chaining, since someone who wants that much probably meant it?, or should we throw a warning?
- I was reading up on some caching done on other parts of the codebase. So something like caching based on `\$past$<depth>$<input_name>$d<stage>` should be fine for the shift register chaining created as a result of $past?

- do we want to support these other global clocking ones too (I'm still trying to figure out how to get explicit clocking events implemented as part of $past's 4th parameter, but just wondering what all would be expected in a single PR):
> 16.9.4 Global clocking past and future sampled value functions
> This subclause describes the system functions available for accessing the nearest past and future values of an expression as sampled by the global clock. They may be used only if global clocking is defined (see 14.14). These functions include the capability to access the sampled value at the global clock tick that immediately precedes or follows the time step at which the function is called. Sampled value is explained in 16.5.1. The following functions are provided.
> Global clocking past sampled value functions are as follows:
> - $past_gclk ( expression )
> - $rose_gclk ( expression )
> - $fell_gclk ( expression )
> - $stable_gclk ( expression )
> - $changed_gclk ( expression )
> Global clocking future sampled value functions are as follows:
> - $future_gclk ( expression )
> - $rising_gclk ( expression )
> - $falling_gclk ( expression )
> - $steady_gclk ( expression )
> - $changing_gclk ( expression )
[...]

PR #237, opened 2025-10-17
WIP: add systemverilog assertions (sva) support to yosys-slang

### LINKED ISSUE #77: SVA (SystemVerilog assertions) support
Support for SVA is missing, e.g. an assertion written in the following way is not supported as input:

```
  assert property (@(posedge clk) req |-> ##3 ack);
```

On the other hand plain assertions like
```
  always @(posedge clk) assert(/property/);
```
are supported.

CheeksTheGeek:
adds a pretty extensive sva converter that can synthesize a bunch of concurrent assertions into actual rtl that yosys can work with. previously yosys-slang was just throwing "unsupported" errors for most sva stuff but now we can actually handle a decent chunk of SVA requirements. I'll need to see how symbiyosys (might not even require) will be modified to handle this, 

resolves #77 
will need to add some work to cover \ #84 

the core architecture:
  - built a new SVAConverter class that lives in sva_converter.{cc,h}
  - uses a three-stage fsm pipeline (nfsm -> ufsm -> dfsm) similar to how verific does it
  - nfsm is the non-deterministic fsm we build while parsing the assertion
  - then we resolve epsilon transitions to get ufsm (unlinked fsm)
  - finally determinize it into dfsm which we can actually synthesize
  - integrated it into slang_frontend so concurrent assertions get routed to the converter instead of erroring out

what's actually supported now:
  - basic assertion statements: assert property, cover property, and expect (treated as procedural assertion)
  - temporal operators: ##n for fixed delays, ##[n:m] for ranged delays (including unbounded ##[n:$])
  - implication operators: |-> (overlapping) and |=> (non-overlapping, which is sugar for |-> ##1)
  - sequence and property declarations - they're templates that get instantiated
  - repetition operators:
    - [*n] consecutive repetition with fixed/ranged counts
    - [=n] non-consecutive repetition
    - [->n] goto repetition
  - sequence operators:
    - throughout - condition must hold throughout a sequence
    - within - sequence must occur within another sequence's window
    - intersect - both sequences must match simultaneously
  - first_match operator for handling multiple match possibilities
  - disable iff for abort conditions on assertions
  - local variables in properties/sequences - they become part of the checker fsm state
  - strong and weak operators (property strength modifiers)
  - abort operators: accept_on, reject_on, sync_accept_on, sync_reject_on
  - conditional properties with if-else
  - case properties for multi-way conditionals
  - sequence methods like .triggered and .matched
  - sva system functions:
    - $past(expr, depth) - historical values using shift registers
    - $rose(signal) - 0->1 transition detector
    - $fell(signal) - 1->0 transition detector
    - $stable(signal) - value unchanged from last cycle
    - $changed(signal) - value changed from last cycle
    - $onehot(vec) and $onehot0(vec) - one-hot checkers using the x&(x-1)==0 trick
    - $isunknown(signal) - always returns 0 in synthesis context
    - $countones(vec) - tree of adders to count set bits
  - multiclock support - each assertion can have its own clocking event
  - wait statements for procedural expect blocks (synthesizes as state machine with waiting flag)
  - zero-length sequences and matches

some implementation details:
  - clocking gets extracted from the assertion's timing control
  - for system functions that need history we create dff chains with the assertion's clock
  - state explosion protection: warns if ctrl signal exceeds 64 bits (sva_fsm_limit)
  - caches condition equality checks to avoid duplicating logic (cond_eq_cache)
  - initial blocks that contain expect/wait get processed as procedural blocks instead of compile-time eval
  - added handling for LocalAssertionVar symbol kind throughout the codebase
  - property and sequence symbols are just skipped during elaboration since they're templates
  - the fsm determinization handles complex sequences with multiple paths efficiently

also did some plumbing work:
  - modified variables.cc to handle LocalAssertionVar symbols
  - updated slang_frontend.cc to route concurrent assertions to the converter
  - handle PropertySymbol and SequenceSymbol declarations properly (skip them, they're templates)
  - extended expression evaluation to handle sva system calls
  - StatementVisitor now has handle() for ConcurrentAssertionStatement and WaitStatement

testing:
  - added 23 test cases in tests/sva/ covering all the major features
  - all 23 tests pass successfully
  - tests cover: temporal delays, implications, sequences, repetitions (consecutive/non-consecutive/goto), throughout/within operators, intersect/first_match, sva functions, property declarations, disable iff, local vars, deferred assertions, expect statements, concurrent assertions, multiclock, complex scenarios, strong/weak, abort operators, conditional properties, case properties, zero length sequences, and sequence methods
  - updated CMakeLists.txt to include the new source files and tests

the fsm approach means we can handle pretty complex temporal properties and the materialization only happens once we've built the full state machine, which keeps the logic reasonably optimized. still have state explosion risks with really complex assertions but that's just how it goes with sva synthesis.

### DIFF
--- a/src/sva_converter.cc
+++ b/src/sva_converter.cc
@@ -277,9 +277,6 @@ void SVAConverter::convert(const ast::ConcurrentAssertionStatement &stmt) {
     // Generate unique name for this assertion
     assertion_name = gen_id("assertion");
 
-    Yosys::log("SVA: ========================================\n");
-    Yosys::log("SVA: Processing assertion in module %s\n", netlist.canvas->name.c_str());
-    Yosys::log("SVA:   propertySpec kind=%d\n", (int)stmt.propertySpec.kind);
     log_debug("SVA: Processing assertion in module %s, propertySpec kind=%d\n",
               netlist.canvas->name.c_str(), (int)stmt.propertySpec.kind);
 
@@ -295,14 +292,12 @@ void SVAConverter::convert(const ast::ConcurrentAssertionStatement &stmt) {
     // Initialize FSM
     reset_fsm();
 
-    // EXACTLY COPY VERIFIC's two-FSM approach (verificsva.cc:1623-1720)
     using AK = ast::AssertionExprKind;
     if (stmt.propertySpec.kind == AK::Binary) {
         auto &bin = stmt.propertySpec.as<ast::BinaryAssertionExpr>();
         using Op = ast::BinaryAssertionOperator;
 
         if (bin.op == Op::OverlappedImplication || bin.op == Op::NonOverlappedImplication) {
-            // Two-FSM approach for implications (Verific lines 1633-1693)
             int node;
 
             // ANTECEDENT FSM
@@ -323,7 +318,7 @@ void SVAConverter::convert(const ast::ConcurrentAssertionStatement &stmt) {
 
             // Fall through to materialize consequent FSM
         } else {
-            // Not an implication, parse normally (Verific lines 1708-1713)
+            // Not an implication, parse normally 
             int node = parse_sequence(stmt.propertySpec, createStartNode());
             createLink(node, acceptNode);
         }
@@ -348,18 +343,15 @@ void SVAConverter::convert(const ast::ConcurrentAssertionStatement &stmt) {
         reject_sig = RTLIL::S0;
     }
 
-    // Enable and assertion signals (Verific lines 1824-1825)
-    // NOTE: Not adding final DFF stage - Verific only adds it conditionally based on clocking.body_net
+    // Enable and assertion signals
+    // NOTE: Not adding final DFF stage - only adds it conditionally based on clocking.body_net
     // For now, use combinational signals directly to match test behavior
     RTLIL::SigBit sig_a = netlist.canvas->Not(NEW_ID, reject_sig);
     RTLIL::SigBit sig_en = netlist.canvas->Or(NEW_ID, accept_sig, reject_sig);
 
-    log("SVA: Assertion %s: sig_a=%s (reject=%s), sig_en=%s\n",
-        log_id(assertion_name), log_signal(sig_a), log_signal(reject_sig), log_signal(sig_en));
-
     RTLIL::SigBit enable_sig = sig_en;
 
-    // Create $check cell (Verific lines 1845-1847)
+    // Create $check cell
     std::string flavor;
     RTLIL::Cell *cell = nullptr;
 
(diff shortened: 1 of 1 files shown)


povik:
Thanks. I wasn't expecting a large drop so soon but I'll take a look.

Can you disclose if LLMs helped write this, and if so, to what extent?

CheeksTheGeek:
Apologies @povik, I just saw this.
And apologies for dropping this out of the blue. 
Spent the weekend trying to get the setup working with symbiyosys's [test setup](https://github.com/YosysHQ/sby/tree/main/tests) (their [comprehensive examples](https://github.com/YosysHQ/sby/tree/main/tests) are also part of the their test flow!
And they all seem to pass!
The first pass was actually kind of a POC from claude code, just had to make a good sized doc to explain how the logic is supposed to work. Your comment on the other issue about verificsva.cc was super helpful for this. 
The first pass seemed to generate good enough FSMs (my first reference of understanding was from Tom Verbeure's blog, such a good read!).
The LLM-generated was not good enough for sby tests though, so had to get down to pen and paper and draw it out, turned out verificsva.cc's two-FSM approach wasn't a parser-convention enforced.

My opinion was that using LLMs to accelerate SW-side dev required for developing platforms for HW can give the same acceleration, but I'm guessing relying solely on LLMs to design huge tasks like this is not a good idea. So, I'm going to cover the whole SVA Converter in a blog (or a discussions' post if you like) with diagrams and all the bells and whistles, just to make sure the whole implementation is covered in theory atleast at one spot, and has been read line by line by me.
This is the reason I created this as a draft PR, to make sure I'm not going down the wrong route (my confidence was not high when starting out) while making sure I'm communicating it's not that ready to start a conversation on.

The sby tests fully passing means I can start covering the under-the-hood details now.

Feel free to test it out by setting the YOSYS_SLANG environment variable and running the [tests through this fork](https://github.com/CheeksTheGeek/sby).

I'm really looking forward to getting an acceptable implementation done at the earliest so there's a real-FOSS implementation of SVA-driven formal out there.
Excited for this 🚀

whitequark:
Using LLMs like this means transferring a huge part of workload that *you* should be doing (ensuring your PR is correct in every way) to the maintainer, who already has plenty of other things to do. This isn't very nice of you.

povik:
I recognize your good intentions but I want to ask you to be very clear and mark any LLM designed code as such in future submissions. I think for me to review a contribution I need there to be a human author which has designed the code and understands every line of it. That doesn't dismiss LLM help altogether (though I'm sceptical) but sets a bar for understanding and control.

Based on your comments and the code I think your PR doesn't meet this bar. It's also too large. Please find ways to split this into smaller PRs. I see e.g. the implementation of `$countones` and friends could be seperated out.

CheeksTheGeek:
I misassumed that the convention for creating Draft PRs is that they're used for early & non-serious reviews, and to show PoCs rather than something merge-ready and serious-review ready.
Will turn this into separate PRs this week. Apologies for the inconvenience.

povik:
I agree with draft PRs being for early work but you're still asking for feedback and you should include a disclaimer for the fact the PR has been extensively LLM generated.

CheeksTheGeek:
kinda serendipitous to see this on hacker news: https://russmiles.substack.com/p/you-cannot-outsource-understanding

Issue #185, opened 2025-06-03
Ignore `wait`


akashlevy:
Not synthesizable, warn user and continue.

CheeksTheGeek:
hi @povik, can I make a PR for [this](https://github.com/CheeksTheGeek/yosys-slang/tree/chore/warn-for-wait-and-expect)?

- wait statement warned about and ignored
- expect warned and ignored about. 
    - given SV LRM gives the relation btw wait's and expect's appearance in code, so similar behaviour should be expected, and since currently SVAs are unsupported they should be dissociated from expect statement in their resulting error/warning
- piling up warn+ignore and flag for program blocks for #177 

no LLMs used btw :)

Issue #177, opened 2025-06-03
`program` should be ignored


akashlevy:
`program` is exclusively for testbenches. This is unimplemented and should be ignored, either by default or with a flag.

PR #239, opened 2025-10-23
Warn & Ignore for wait, expect and program

CheeksTheGeek:
Covers #185 
Covers #177 

- wait statement warned about and ignored
- expect warned and ignored about. 
    - given SV LRM gives the relation btw wait's and expect's appearance in code, so similar behaviour should be expected, and since currently SVAs are unsupported they should be dissociated from expect statement in their resulting error/warning
- program blocks warned and ignored about with a flag

no LLMs used

### DIFF
--- a/src/slang_frontend.cc
+++ b/src/slang_frontend.cc
@@ -1113,7 +1113,6 @@ public:
 	void handle(const ast::WaitStatement &stmt)
 	{
 		netlist.add_diag(diag::WaitStatementUnsupported, stmt.sourceRange);
-		stmt.stmt.visit(*this);
 	}
 
 	void handle(const ast::Statement &stmt)
--- a/tests/various/expect_test.ys
+++ b/tests/various/expect_test.ys
@@ -9,5 +9,4 @@ module expect_test(input logic clk, input logic a, input logic b, input logic c,
 endmodule
 EOF
 
-design -save before
-design -load before
+test_slangdiag -expect "expect statement will not be synthesized"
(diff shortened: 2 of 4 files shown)


povik:
> Also, ./tests/run.sh seems to not give good colored outputs because of use of \e rather than \033[ because the former's is only limited to bash and is not posix-compliant [..] would it be fine if I also add that in this PR?

Yes, that would be good to fix and feel free to add it here

src/initial_eval.cc
```diff
@@ -706,6 +706,13 @@ ER EvalVisitor::visit(const TimedStatement &stmt)
 	return result;
 }
 
+ER EvalVisitor::visit(const slang::ast::WaitStatement &stmt)
+{
+	context.addDiag(slang_frontend::diag::WaitStatementUnsupported, stmt.sourceRange);
+	ER result = stmt.stmt.visit(*this);
```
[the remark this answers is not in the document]
CheeksTheGeek: ah, understood

src/diag.cc
```diff
@@ -95,19 +96,21 @@ namespace diag {
 	DiagCode PortCorrespondence(DiagSubsystem::Netlist, 1042);
 	DiagCode UnsynthesizableFeature(DiagSubsystem::Netlist, 1043);
 	DiagCode SVAUnsupported(DiagSubsystem::Netlist, 1044);
-	DiagCode ForbiddenDemotion(DiagSubsystem::Netlist, 1045);
-	DiagCode UdpUnsupported(DiagSubsystem::Netlist, 1046);
```
povik: Next time please add new diagnostics to the end so that it's clear nothing else changed

CheeksTheGeek: understood

PR #245, opened 2025-11-10
Add conditional support of Dual Edge FFs

### LINKED ISSUE #164: Support for dual-edge FFs `always (@edge clk)`
Thought we should make an issue for this. Can synthesize this using a trick: https://vlsi-soc.blogspot.com/2013/06/dual-edge-triggered-flip-flop.html

CheeksTheGeek:
Resolves #164 

- implements the first pattern from https://vlsi-soc.blogspot.com/2013/06/dual-edge-triggered-flip-flop.html 
    - is only allowed on explicit allowance based on `--allow-dual-edge-ff`
    - 2 FFs + mux
    - simply put, (posedge ff --> intermediary --> negedge FF --> intermediary) mux selects on these based on clock
    - also an async reset impl for this
    - Made sure to use helpers, similar to canvas->addDff and canvas->addAldff for the dual edge versions
 
cc: @povik @whitequark

### DIFF
--- a/src/builder.cc
+++ b/src/builder.cc
@@ -5,6 +5,7 @@
 // Distributed under the terms of the ISC license, see LICENSE
 //
 #include "slang_frontend.h"
+#include "variables.h"
 
 namespace slang_frontend {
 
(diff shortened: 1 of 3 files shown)


povik (changes requested):
See comment; I've also pushed a functional test to your branch

PR #254, opened 2025-11-27
RFC, Add implicit synthesis define

Emin017:
* Added a `--no-synthesis` flag to the command-line interface, allowing users to prevent the implicit addition of `-D SYNTHESIS`.

Adding this feature ensures consistency with the default behavior of the [read_verilog](https://yosyshq.readthedocs.io/projects/yosys/en/latest/cmd/index_frontends.html#cmd-read_verilog) command in Yosys. Additionally, languages like [Chisel](https://www.chisel-lang.org/) often generate Verilog code with a `SYNTHESIS` parameter defined to maintain the synthesizability of the code. The implicit definition of this parameter can help optimize the user experience by automatically ensuring the code is suitable for synthesis. Furthermore, we provide the `--no-synthesis` option, allowing users to override this implicit definition when desired.


whitequark (approved):

povik (changes requested):
I agree with adding the define; only two tweaks to make

PR #268, opened 2026-01-06
Improve cell naming in assertion handling by using labels if available

gipsyh:
When using immediate assertions with labels (e.g., ast: assert(cand);), the label name was not being preserved in the generated $check cell name. Instead, an auto-generated name like $13 was used, causing the label to be lost in downstream BTOR/YWB exports.


gipsyh:
> @gipsyh I've found a way to implement the naming. Your tests should be passing. I'll merge once the CI is green.

Ok, thanks a lot! 😊

PR #278, opened 2026-01-26
Implement foreach loop support feature

likeamahoney:
Hello, everyone! I have added support for unrolling and gererating RTLIL for `foreach` loops and included corresponding tests.

In most cases, these loops are unrollable, although it is possible that I have overlooked some scenarios. There are, of course, more complex cases, but I aimed to implement basic support that works for the general case.

### DIFF
--- a/src/statements.h
+++ b/src/statements.h
@@ -558,26 +558,29 @@ public:
 
 	void handle(const ast::ForeachLoopStatement &stmt)
 	{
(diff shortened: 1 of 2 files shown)


povik:
> There are, of course, more complex cases, but I aimed to implement basic support that works for the general case.

@likeamahoney I wonder about these more complex cases. Can you think of some which will be incorrectly processed by the now merged support?

PR #285, opened 2026-02-02
Support unnamed port connections for blackboxes

likeamahoney:
Hi everyone,
quite often in open-source designs you can find blackbox module instances without named port connections. I decided to improve the support for this case, added a corresponding test, and slightly relaxed the diagnostics to only forbid complex expressions for now.

### DIFF
--- a/src/diag.cc
+++ b/src/diag.cc
@@ -89,7 +89,6 @@ DiagCode FixedSizeRequired(DiagSubsystem::Netlist, 1033);
 DiagCode AloadOne(DiagSubsystem::Netlist, 1034);
 DiagCode BadInlinedPortConnection(DiagSubsystem::Netlist, 1035);
 DiagCode NoParamsOnUnkBboxes(DiagSubsystem::Netlist, 1037);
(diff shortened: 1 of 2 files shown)


src/slang_frontend.cc
```diff
@@ -2265,13 +2265,28 @@ struct PopulateNetlist : public TimingPatternInterpretor, public ast::ASTVisitor
...
+
+				if (port_name.empty()) {
+					netlist.add_diag(diag::SimpleConnNameRequiredOnUnkBboxes, sym.location);
+					continue;
+				}
 			}
 
-			auto &expr = port_conns[i]->as<ast::SimpleAssertionExpr>().expr;
-			cell->setPort(RTLIL::escape_id(std::string{port_names[i]}), netlist.eval(expr));
+			cell->setPort(netlist.new_id(), netlist.eval(expr));
```
povik: @likeamahoney we are now always setting placeholder names even if the AST has a non-empty port name

likeamahoney: Sorry about that — fixed in the latest commit. Now the predefined port name is set if present.

PR #297, opened 2026-03-25
Add clog2 system call lowering

likeamahoney:
Adds support for lowering the `$clog2` system function to RTLIL.
                                                                                                                                                                                                                                                       The implementation computes `$clog2(n)` as the MSB position of `n-1` plus 1, using a priority encoder built from a chain of multiplexers. The `n=0` edge case (where `n-1` wraps to all-ones) is handled by gating the output with `$reduce_bool(n)`.

### DIFF
--- a/src/slang_frontend.cc
+++ b/src/slang_frontend.cc
@@ -1229,6 +1229,11 @@ RTLIL::SigSpec EvalContext::operator()(ast::Expression const &expr)
 					auto arg = call.arguments()[0];
 					auto sig = (*this)(*arg);
 					ret = netlist.CountOnes(sig, (int)call.type->getBitstreamWidth());
+				} else if (name == "$clog2") {
+					require(expr, call.arguments().size() == 1);
+					auto arg = call.arguments()[0];
+					auto sig = (*this)(*arg);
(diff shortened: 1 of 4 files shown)


src/slang_frontend.cc
```diff
@@ -1229,6 +1229,11 @@ RTLIL::SigSpec EvalContext::operator()(ast::Expression const &expr)
 					auto arg = call.arguments()[0];
 					auto sig = (*this)(*arg);
 					ret = netlist.CountOnes(sig, (int)call.type->getBitstreamWidth());
+				} else if (name == "$clog2") {
+					require(expr, call.arguments().size() == 1);
```
povik: Does slang enforce the number of arguments to 1? If so, we should use `ast_invariant`

povik (commented):
One comment otherwise looks fine

I'm in the middle of a large refactoring and it might take me a week or longer to merge this

povik (approved):
Thanks!

src/slang_frontend.cc
likeamahoney: Yes, slang enforces exactly one argument here, so ast_invariant is the right choice - require isn't needed. I've already fixed this in the latest commit. Thanks!

Issue #243, opened 2025-11-01
Inference of multi-dimensional memories


povik:
Take the following example
```
module top(input logic [3:0] a1, input logic [3:0] a2, input logic clk,
			output logic [7:0] y, input logic [7:0] d);
	logic [7:0] array[16][16];
	always @(posedge clk) begin
		array[a1][a2] <= d;
		y <= array[a1][a2];
	end
endmodule
```
yosys-slang infers memory addressing on the final unpacked dimension only, i.e. the memory width is 128. Yosys proper infers a memory of width 8, as does another tool. This ticket is to add the same behavior either as the default or an option.

PR #298, opened 2026-03-28
Support $readmemh() and $readmemb()

### LINKED ISSUE #126: Unable to initialize ROM memory
As the title states, I am unable to load a memory with initial values, be it a rom or ram. 
Then usage of `initial $readmemh(init_file, memory)` doesn't seem to work.
I am using the following memories, [bram_sdp](https://github.com/projf/projf-explore/blob/main/lib/memory/bram_sdp.sv) and [rom_sync](https://github.com/projf/projf-explore/blob/main/lib/memory/rom_sync.sv).

I load these memories with `read_slang --allow-dup-initial-drivers bram_sdp.sv`, The extra `-allow-dup-initial-drivers` is needed because otherwise yosys throws an error that the memory is multi driven.

I had also tried adding the attributes `(* ram_style = "block", ram_init_file = "mem/init_file.mem" *)` but that didn't do anything. I didn't see anything in the logs.

I am building for the ECP5 fpga's (ulx3s board). This is how my full build looks like:
```
  $(YOSYS) -l build/yosys.log \                                                                                                                                                                                                   
    »       »       -m slang \                                                                                                                                                                                                      
    »       »       -f slang \                                                                                                                                                                                                      
    »       »       -p "read_verilog $(CLK_SOURCES) $(V_BLACKBOX)" \                                                                                                                                                                
    »       »       -p "read_slang --empty-blackboxes \                                                                                                                                                                             
    »       »       -I src/ -I src/memory -I build/clocks \                                                                                                                                                                         
    »       »       --top top --allow-dup-initial-drivers  \                                                                                                                                                                        
    »       »        $(SV_MEMORY) $(SV_SOURCES) top.sv \                                                                                                                                                                            
    »       »       --keep-hierarchy " \                                                                                                                                                                                            
    »       »       -p 'synth_ecp5 -json $(JSON)                 
```
I had also tried loading the memories with just `read -sv $(SV_MEMORIES)` but then I get an error that the memory instantiations are not found, even if I add the flags `--allow-use-before-declare --ignore-unknown-modules` it didn't work. I had even tried the `--extern-modules` flag, but that did nothing. I've tried reordering the read commands, but that also didn't get me far.

max-kudinov:
Fixes #126.

This is a draft as I want to clarify a few things.

I mostly adapted how Yosys [handles](https://github.com/YosysHQ/yosys/blob/74fef31bf2955c1b57cabf0eeb65543854a7c72f/frontends/ast/simplify.cc#L4691) parsing memory files and emitting `$meminit_v2` cells. Additional checks and logic was added to match IEEE 1800-2023 21.4.

The logic is the following:

With `$readmem[h/b]` we get from 2 to 4 arguments. 1) File name to read the data from 2) Memory symbol to place the data into 3) Start memory address 4) Finish memory address.

We evaluate arguments 1, 3 and 4 to check that they are elaboration time constants. No checks on memory, because slang already ensures that it is an unpacked array.

Then we check with `filename_arg->isImplicitString()` that filename is either a string or could be evaluated to string.

There are 2 ways a file name can be passed:

1. Explicitly: `$readmemh("file.hex", mem)`
2. Through a parameter: `$readmemh(FILE_PARAM, mem)`

Then we try to open a file. There's no specification in IEEE if the path is absolute or relative to something, I've done what yosys does: it tries to open a file relative to current executable directory, and if it fails, it tries to open it relative to the file containing `$readmem` function.

Then we get start and finish addresses if specified. What is important, is that those addresses represent only highest of dimensions, so for example `logic [31:0] mem [2:7][0:3][5:8]` the valid address range is from 2 to 7.

We read a file line by line, ignoring comments and splitting a line into tokens, which either represent data word or an address for a jump. Addresses start with `@` followed by a hex value.

We read one word at a time, convert it to `std::vector <RTLIL::State`, then expand or truncate if the width of a word doesn't match the width of memory element. We accumulate data in that vector until we have to jump addresses, in that case new $meminit_v2 cell is emitted, vector clears and everything starts over with new address. When start address is higher then finish address argument, then we emit new cell for each element. All of this matches Yosys behavior.

After we're done filling memory in the specified range, we still continue reading the file until the end to get the total number of words and print a warning if the number of words differ from the range as per IEEE.

IEEE states that the start and finish addresses are for the outer unpacked dimension. So if we have multiple unpacked dimensions, we flatten them to a single range and then we'll have to scale all addresses from the arguments and from the file. This is not done in original frontend.

A few things to clarify:

1) Yosys frontend has a weird behavior when the unpacked memory range doesn't start from 0, for example `logic [31:0] mem [4:8]`. In this case the range would start from 4 and the first `ADDR` in `$meminit_v2` would be also 4. In this case yosys-slang also does this because of `mem->start_offset` in `void RTLILBuilder::emit_meminit_cell`. However, if memory has multiple unpacked dimensions, regardless of lower index in highest dimension, Yosys frontend always starts the range from 0. For example, `logic [31:0] mem [5:7][1:3]` would result in a range from 0 to 8. 

In this case we have a mismatch, because yosys-slang still accounts `start_offset`, but Yosys no longer does. I'm not sure if this is intentional and which behavior is correct.

2) There was a problem with size and width of `RTLIL::Memory` in `add_internal_wires`. Current code assumes that there is only 1 unpacked dimension, so it writes the range of outer dimension as size, and the rest as width. I changed it to go through all unpacked dimensions and calculate total size, and the width is everything that's left (unpacked size). I haven't tested it for regressions yet, but it does the trick for my case.

3) `$memrd_v2` cell has the same problem, current code grabs only outer `ast::ElementSelectExpression` and calculates `WIDTH`, `ADDR` and `ABITS` based on only 1 selector. I'm not sure how to solve this, as we have to merge multiple `SigSpec` selectors into single `ADDR`.

4) IEEE states that `$readmem` functions could be called during any time of simulation, not just in elaboration time. So Yosys provides an option to do this:

```SystemVerilog
always_ff @(posedge clk)
    $readmemh("memfile.bin", mem);
```

In that case, a bunch of assignments are created inside of a process. I'm not sure how to approach this though, so for now my code assumes that the function is called from `initial`.

5) So far I've been manually comparing Yosys RTLIL with yosys-slang. Not sure how to write tests for this, as it requires reading an external file.


povik:
> 2. There was a problem with size and width of RTLIL::Memory in add_internal_wires. Current code assumes that there is only 1 unpacked dimension, so it writes the range of outer dimension as size, and the rest as width. I changed it to go through all unpacked dimensions and calculate total size, and the width is everything that's left (unpacked size). I haven't tested it for regressions yet, but it does the trick for my case.

On this point: Yes, currently the tool infers memory addressing on the outermost unpacked dimension only (see ticket #243). Are you running into this limitation with real world designs? I think the change to add_internal_wires is inconsistent unless we adjust the emission of `$memwr_v2` and `$memrd_v2` too. In any case let's leave resolving #243 for another PR and separate it from readmemh/b

povik:
I'll push to your branch to bring it up to speed with master

max-kudinov:
> Are you running into this limitation with real world designs?

Yes, I discovered it when testing this PR on my own project. I use multiple lookup tables, where each table is a matrix, so I had memory like this: `logic [W_PIXEL-1:0] recode_lut [NUM_TEXTURES] [NUM_COLORS]`. It simplifies addressing in SV a little bit. Default Yosys frontend works fine with it.

> I think the change to add_internal_wires is inconsistent unless we adjust the emission of $memwr_v2 and $memrd_v2 too. In any case let's leave resolving https://github.com/povik/yosys-slang/issues/243 for another PR and separate it from readmemh/b

Yeah, initially I wanted to change memory addressing here as well, but I just wanted your thoughts on how to approach it. But leaving this for another PR seems better, I agree.

src/slang_frontend.cc
```diff
@@ -801,6 +801,288 @@ RTLIL::SigSpec handle_past(EvalContext &eval, const ast::CallExpression &call)
...
+	for (int i = orig.size() - width; i >= 0; i -= width)
+		bits.insert(bits.end(), orig.begin() + i, orig.begin() + i + width);
+	return bits;
+}
+
+/*
+ * The following two functions contain parts adapted from Yosys
+ * Copyright (C) 2012  Claire Xenia Wolf <claire@yosyshq.com>
+ * Licensed under the ISC License
+ */
```
povik, line 818: I saw some similarities with yosys so it doesn't harm to reproduce its notice

src/slang_frontend.cc
```diff
@@ -801,6 +801,288 @@ RTLIL::SigSpec handle_past(EvalContext &eval, const ast::CallExpression &call)
...
+			const_ = reverse_data(const_, word_size);
+			hdl_base = cursor;
+		} else {
+			hdl_base = cursor - (const_.size() / word_size) + 1;
+		}
+
+		uint32_t base = target_range.isLittleEndian() ? hdl_base - target_range.right
+													  : target_range.right - hdl_base;
+
+		context.do_simple_assign(call.sourceRange.start(),
```
povik, line 994: The data is now comitted using `do_simple_assign` instead of emitting `$meminit_v2` directly. This means it will work on non-memories with unpacked dimensions (under `--no-implicit-memories` or if inference is prevented for another cause) and it will also resolve your question 4 by handling occurrences in always_ff

The meminit cell gets emitted anyway it just happens inside the helper

povik:
Ad 1, it's not important to match yosys behavior as long as the synthesized behavior is correct which seems to be the case with present yosys-slang handling.

2, 3 left for later.

4 is dealt with by the change to use `do_simple_assign`.

Ad 5, we'll need a few tests before merging. The extra file is not an issue. The testing can be similar to `meminit.ys` in that we compare a memory initialized using readmemh to one initialized by other means. I'll see if I can help with testing later.

src/slang_frontend.cc
```diff
@@ -801,6 +801,288 @@ RTLIL::SigSpec handle_past(EvalContext &eval, const ast::CallExpression &call)
...
+				/* blocking= */ true);
+		data.clear();
+	};
+
+	while (!f.eof()) {
+		std::string line, token;
+		std::getline(f, line);
+
+		// Remove multiline comments
+		bool in_comment = false;
```
max-kudinov: Multiline comment state flag has to persist between lines, so it has to be outside line reading loop

max-kudinov:
For some reason I thought that your test system wont handle external files, I should've actually taken a look a long ago. Anyway, I was just writing tests too, perhaps having a case with X and Z values in the memory init file would be a good idea in addition to what you've already added.

povik:
> perhaps having a case with X and Z values in the memory init file would be a good idea in addition to what you've already added.

I just saw in the [codecov report](https://app.codecov.io/gh/povik/yosys-slang/pull/298) that the X/Z paths are not getting exercised so additional tests would be good. I'll wait for any tests you want to add, other than that the PR looks good to merge.

max-kudinov:
Sat assertions fail when a memory with X or Z elements is compared to a constant, perhaps some evaluation differences. So I've added a simple check without `(* rom_block *)` attribute, seems to be sufficient

povik:
Thanks for the contribution. I'm glad we filled this long-standing gap. 🎉

PR #299, opened 2026-04-12
[bug-fix] Fix big-endian unpacked array indexing

likeamahoney:
This patch corrects four related bugs in how big-endian (BE) ranges are handled throughout the frontend.

Problem:
SystemVerilog allows both big-endian ([0:N]) and little-endian ([N:0]) range declarations for packed vectors and unpacked arrays. Several code paths in the frontend mishandled BE ranges, producing wrong synthesis output:

1. AddressingResolver treated BE unpacked arrays like BE packed vectors. Unpacked arrays are always stored sequentially by logical index regardless of declaration order, but interpret_index was applying the bit-reversal transformation intended for packed BE vectors, making arr[i] silently access the wrong element under `--no-implicit-memories`.
2. `interpret_index` used zero-extension before NOT, breaking negative indices. For packed vectors with negative index ranges (e.g. `logic [-7:-2]`), the zero-extended value is large and positive, and Not produces a wildly wrong shift amount, yielding x for all in-range accesses.
3. Constant aggregate evaluation iterated BE unpacked arrays in the wrong order. `convert_const` used rbegin for all unpacked arrays. After normalising BE ranges to LE in `AddressingResolver`, element `[lower]` must be at the LSB, but reversing a BE array's declaration order (which already starts at the lowest index) puts the highest-index element there instead.
4. `SimpleAssignmentPattern` on both RHS and LHS handled BE arrays incorrectly. The synthesis-path evaluator and the procedural LHS decomposer both iterated elements in the wrong order for BE unpacked arrays, causing mismatches between constant folding and synthesis results.

Fix:
  - `src/addressing.cc`: Normalise BE unpacked array ranges to LE in both `ElementSelectExpression` and `RangeSelectExpression` constructors, so `interpret_index` always sees a LE range for unpacked arrays. Replace zero-extension with sign-extension in `interpret_index`s BE branch (1{signal.msb(), signal}` instead of `{S0, signal}`), making the formula correct for all signed index values.
  - `src/slang_frontend.cc`: Add an optional type parameter to `convert_const` - use forward element iteration for BE unpacked arrays so the physical layout matches the LE-normalised addressing. Reverse `SimpleAssignmentPattern` element iteration for BE unpacked arrays on the RHS synthesis path. 
  - `src/procedural.cc`: Reverse `SimpleAssignmentPattern` element iteration for BE unpacked arrays on the LHS decomposition path. Remove the now-redundant `big_endian`
  -  flag from `add_memory_init` - `convert_const` already produces element `[i]` at physical position (`i` − lower) for all arrays.

### DIFF
--- a/tests/various/be_unpacked_array.ys
+++ b/tests/various/be_unpacked_array.ys
@@ -1,0 +1,185 @@
+# Test big-endian unpacked arrays ([0:N] range) are addressed correctly.
+#
+# With --no-implicit-memories the array goes through AddressingResolver instead
+# of the $mem path.  A bug caused the resolver to reverse the index for BE
+# unpacked arrays (as if they were packed BE vectors), making arr[i] access
+# element (range.right - i) instead of element i.
+
+# [0:3] and [3:0] unpacked arrays must be equivalent under --no-implicit-memories.
+# Uses equiv_make/induct so shared state is matched and proof is fast.
+design -reset
+read_slang --no-implicit-memories <<EOF
+module be_top(input clk, input [1:0] addr, input din, output dout);
+    reg arr [0:3];
+    always @(posedge clk)
+        arr[addr] <= din;
+    assign dout = arr[addr];
+endmodule
+EOF
+proc; opt_clean
+
+read_slang --no-implicit-memories <<EOF
+module le_top(input clk, input [1:0] addr, input din, output dout);
+    reg arr [3:0];
+    always @(posedge clk)
+        arr[addr] <= din;
+    assign dout = arr[addr];
+endmodule
+EOF
+proc; opt_clean
+
+equiv_make le_top be_top equiv_top
+equiv_induct equiv_top
+equiv_status -assert
+
+# Same equivalence must hold with the default (memory-inferred) path.
+design -reset
+read_slang <<EOF
+module be_top(input clk, input [1:0] addr, input din, output dout);
+    reg arr [0:3];
+    always @(posedge clk)
+        arr[addr] <= din;
+    assign dout = arr[addr];
+endmodule
+EOF
+proc; memory; memory_map; opt_clean
+
+read_slang <<EOF
+module le_top(input clk, input [1:0] addr, input din, output dout);
+    reg arr [3:0];
+    always @(posedge clk)
+        arr[addr] <= din;
+    assign dout = arr[addr];
+endmodule
+EOF
+proc; memory; memory_map; opt_clean
+
+equiv_make le_top be_top equiv_top
+equiv_induct equiv_top
+equiv_status -assert
+
+# BE packed vector [0:3]: dynamic bit-select must return the correct bit.
+# v[addr] on logic [0:3] should equal v_le[3-addr] of its LE mirror.
+# Tests the zero-extension fix in interpret_index.
+design -reset
+read_slang <<EOF
+module top(input [1:0] addr, input [0:3] v,
+           output bit_be, output bit_le);
+    assign bit_be = v[addr];
+    logic [3:0] v_le;
+    assign v_le = {v[0], v[1], v[2], v[3]};
+    assign bit_le = v_le[3 - addr];
+    always_comb
+        assert(bit_be === bit_le);
+endmodule
+EOF
+prep
+chformal -lower
+sat -verify -enable_undef -prove-asserts
+
+# BE vs LE unpacked array with non-zero lower bound: [2:5] vs [5:2].
+# Both arrays must return the same element for the same logical index.
+design -reset
+read_slang --no-implicit-memories <<EOF
+module be_top(input clk, input [1:0] idx, input din, output dout);
+    reg arr [2:5];
+    always @(posedge clk)
+        arr[idx + 2] <= din;
+    assign dout = arr[idx + 2];
+endmodule
+EOF
+proc; opt_clean
+
+read_slang --no-implicit-memories <<EOF
+module le_top(input clk, input [1:0] idx, input din, output dout);
+    reg arr [5:2];
+    always @(posedge clk)
+        arr[idx + 2] <= din;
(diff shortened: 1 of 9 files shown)


povik:
I'm not convinced there's a real bug but I'm open to more convincing.

I have tried the tests from this PR on master and the sole failing one is be_unpacked_array.ys. I think it fails because it makes the assumption that the internal layout of `arr` matches between the LE and BE cases.

My understanding is the LRM does not specify the internal layout of unpacked arrays. yosys-slang uses this freedom and treats unpacked and packed arrays the same.

tests/various/be_unpacked_array.ys
```diff
@@ -0,0 +1,185 @@
...
+module le_top(input clk, input [1:0] addr, input din, output dout);
+    reg arr [3:0];
+    always @(posedge clk)
+        arr[addr] <= din;
+    assign dout = arr[addr];
+endmodule
+EOF
+proc; opt_clean
+
+equiv_make le_top be_top equiv_top
```
povik, line 31: `equiv_make` will assume the serialized view (layout) of `arr` equivalent between `le_top` and `be_top`

likeamahoney: You're right about `equiv_make`. If master stores arr[k] at physical bit N−k for BE arrays (treating them like BE packed), then `equiv_make` between `le_top` and `be_top` aligns their state bits by position which is wrong and the proof fails vacuously. That test is indeed misleading and I can remove it.

But tests/various/meminit.ys (which the PR modifies) shows a concrete bug that has nothing to do with state alignment. Consider this module:

```
(* rom_block *) reg [3:0] x[0:1];
                reg [3:0] y[0:1];
initial begin x = '{7, 1}; y = '{7, 1}; end
always_comb assert(x[0] === y[0] && x[1] === y[1]);
```

Both x and y get the same initializer `'{7,1}`, meaning `arr[0]=7, arr[1]=1`. After memory_collect+memory_map, x becomes a ROM read combinationally, y stays as flops addressed dynamically. The SAT check fails on master - x[0] returns 1 and y[0] returns 7. The values are swapped.

The cause: convert_const iterates elements with rbegin for all unpacked arrays, so for BE [0:1] the INIT constant has arr[1]'s value at bit 0 and arr[0]'s value at bit 4. The `$mem` path reads address 0 and gets `arr[1]`'s value. The --no-implicit-memories flop path, on the other hand, applies Not(index) to map address 0 to physical bit 4 and gets `arr[0]`'s value correctly. Two paths in the same module, same initializer, inconsistent results. That's a bug regardless of what internal layout you choose.

likeamahoney:
> I'm not convinced there's a real bug but I'm open to more convincing.
> 
> I have tried the tests from this PR on master and the sole failing one is be_unpacked_array.ys. I think it fails because it makes the assumption that the internal layout of `arr` matches between the LE and BE cases.
> 
> My understanding is the LRM does not specify the internal layout of unpacked arrays. yosys-slang uses this freedom and treats unpacked and packed arrays the same.



> I'm not convinced there's a real bug but I'm open to more convincing.
> 
> I have tried the tests from this PR on master and the sole failing one is be_unpacked_array.ys. I think it fails because it makes the assumption that the internal layout of `arr` matches between the LE and BE cases.
> 
> My understanding is the LRM does not specify the internal layout of unpacked arrays. yosys-slang uses this freedom and treats unpacked and packed arrays the same.

The issue isn't which internal layout is chosen - it's that master uses inconsistent layouts for the same array in different code paths.

Two parts of yosys-slang handle BE unpacked arrays:
1. `convert_const` - converts aggregate literals like `arr = '{7, 1}` to a bit constant (used for ROM init, initial blocks, etc.)
2. `AddressingResolver` - maps a dynamic index like `arr[i]` to a physical bit range

On master, convert_const iterates elements with rbegin for all unpacked arrays, so for arr `[0:1]` with `'{7, 1}` it produces a constant where `arr[1]=1` is at bits `[3:0]` and `arr[0]=7` is at `bits [7:4]`. Meanwhile, `AddressingResolver` independently maps index 0 to bit position 4 and returns `arr[0]=7`.

So dynamic reads are internally consistent with each other - a dynamic write to arr[0] followed by a dynamic read from arr[0] gives the right value. But as soon  as there's a constant initializer, the layout convert_const uses doesn't match what AddressingResolver expects. Write arr = `'{7, 1}` (via initial), then read
`arr[0]` dynamically: you get 1, not 7.

The LRM is unambiguous on this point: after `arr = '{7, 1}`, `arr[0]` must equal 7. That's not a layout question - it's a correctness requirement.

The fix makes `convert_const` and `AddressingResolver` agree on a single convention for BE unpacked arrays. The layout itself doesn't matter; the consistency does.

povik:
Will you help me by providing a reproducer? I believe I tried meminit.ys from this PR and it passed on master.

likeamahoney:
> Will you help me by providing a reproducer? I believe I tried meminit.ys from this PR and it passed on master.

You're right. I was testing against a non-stock, broken yosys build - that's why I was seeing the LEC failure in the first place. When I checked against stock `yosys`, everything passes correctly.

After more careful testing against actual master, I can confirm there is no semantic bug. Both `$mem` and `--no-implicit-memories` paths handle BE unpacked arrays correctly - they use a reversed physical layout (same as BE packed), and that convention is applied consistently everywhere, so read-after-write is correct. The `be_unpacked_array.ys` test fails on master exactly for the reason you described - `equiv_make` assumes the same state encoding between le_top and be_top, which isn't guaranteed when the two use different layout conventions.

The PR's actual effect is changing the internal layout of BE unpacked arrays from the "reversed" convention to the  forward" (LE-aligned) convention. That's an internal consistency improvement, not a bugfix.

I can close this PR. Thank you for pushing back and asking for a proper reproducer - it forced a much more careful analysis.

povik:
I'm glad we were able to resolve it.

PR #300, opened 2026-04-20
Fix interface output port undriven detection in nested --keep-hierarchy

likeamahoney:
When a non-dissolved module (`foo`) contained another non-dissolved instance (`inst2`) driving an interface output port, `register_driven` was not called for that port. As a result, `finalize_variable_initialization` connected the output to `X`, and subsequent opt passes removed the driving instance entirely.
                                                                                                                                                                                                                                                      
   
**Fix**: call `netlist.register_driven(port)` in the scopes_remap branch for Out/InOut modport ports.


povik:
Thanks, this is a severe issue you have fixed. I'll wait for the CI to show green after my changes and then merge

PR #303, opened 2026-04-23
 Support assignment patterns on the LHS of port connections

phsauter:
The test added failed before but is properly handled now.

Each port connection to an input or output port is a continuous assignment from source to sink (23.3.3). Simple assignment patterns on the left-hand side of such assignments were rejected, e.g.:
    .y_o ('{y_o[0], y_o[1]})

Lower them element-wise to concatenation LValues, matching the existing procedural handling.

### DIFF
--- a/src/lvalue.cc
+++ b/src/lvalue.cc
@@ -7,7 +7,6 @@
 #include "slang/ast/expressions/ConversionExpression.h"
 #include "slang/ast/expressions/MiscExpressions.h"
 #include "slang/ast/expressions/OperatorExpressions.h"
-#include "slang/ast/expressions/AssignmentExpressions.h"
 #include "slang/ast/expressions/SelectExpressions.h"
 #include "slang/ast/symbols/MemberSymbols.h"
 #include "slang/ast/symbols/VariableSymbols.h"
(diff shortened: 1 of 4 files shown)


phsauter:
> ❌ Patch coverage is `76.92308%` with `3 lines` in your changes missing coverage. Please review.

Not sure how I can cover the remaining line mr codecov bot.

src/lvalue.cc
```diff
@@ -112,6 +113,20 @@ std::optional<LValue> LValue::analyze(
...
+	case ast::ExpressionKind::SimpleAssignmentPattern: {
+		std::vector<LValue> elements;
+		const auto &pattern = expr.as<ast::SimpleAssignmentPatternExpression>();
+		for (auto element : pattern.elements()) {
+			ast_invariant(expr, element->kind == ast::ExpressionKind::Assignment);
+			auto &assign = element->as<ast::AssignmentExpression>();
+			auto element_lv = analyze(context, assign.left(), silent);
+			if (!element_lv)
+				return std::nullopt;
+			elements.push_back(std::move(*element_lv));
```
povik: You are assuming no conversion inside `assign.right()`. The test I will push has a case with conversion

povik:
With more planning assignment patterns could be handled in lvalue.cc for all contexts (continuous, procedural, and driver ports). Right now it's easier to have separate handling for the three contexts. I don't mind the duplication

PR #305, opened 2026-04-24
Add configurable ff naming schemes

phsauter:
The legacy scheme remains the default and keeps the existing `$driver$<signal><subfield>` naming.

Added options:
- `--ff-naming=legacy|signal|auto` Selects the naming scheme for inferred flip-flops. `signal` derives names only from the driven signal and subfields, while `auto` uses named always_ff blocks and after that the signal-based scheme.
- `--ff-prefix=<prefix>` Prepends a custom prefix to inferred flip-flop names. By default this is empty for the new schemes and `$driver$` for `legacy`.
- `--ff-suffix=<suffix>` Appends a custom suffix to non-legacy flip-flop names. The default is `_reg`, the suffix is not duplicated if the base name already ends in it.

It also flattens struct and generate-block hierarchy into the name, distinguishes scope numbers from array indices and shows packed slices.

Some example names it generates:
```
gpio$croc_chip.i_croc_soc.i_croc.i_gpio/gen_gpios_1_serial_q_reg
gpio$croc_chip.i_croc_soc.i_croc.i_gpio/gen_gpios_0_serial_q_reg
soc_ctrl_regs$croc_chip.i_croc_soc.i_croc.i_soc_ctrl/core_status_q_reg
cve2_if_stage$croc_chip.i_croc_soc.i_croc.i_core_wrap.i_core.if_stage_i/instr_fetch_err_o_reg
cve2_if_stage$croc_chip.i_croc_soc.i_croc.i_core_wrap.i_core.if_stage_i/instr_is_compressed_id_o_reg
cve2_if_stage$croc_chip.i_croc_soc.i_croc.i_core_wrap.i_core.if_stage_i/instr_rdata_c_id_o_reg
cve2_if_stage$croc_chip.i_croc_soc.i_croc.i_core_wrap.i_core.if_stage_i/instr_rdata_alu_id_o_reg
fifo_v3$croc_chip.i_croc_soc.i_croc.i_uart.i_uart_rx.i_fifo_v3/mem_q_2_reg
fifo_v3$croc_chip.i_croc_soc.i_croc.i_uart.i_uart_rx.i_fifo_v3/mem_q_1_reg
fifo_v3$croc_chip.i_croc_soc.i_croc.i_uart.i_uart_rx.i_fifo_v3/mem_q_0_reg
```

### DIFF
--- a/src/naming.cc
+++ b/src/naming.cc
@@ -81,7 +81,8 @@ std::vector<NamedChunk> generate_subfield_names(VariableChunk chunk, const ast::
 	return ret;
 }
 
-// Format one raw hierarchy or signal name into a friendly identifier
+// Format one raw hierarchy or signal name into an alphanumeric+underscore identifier fragment
+// Makes it easy to use and compose in Tcl-style contexts without escaping characters
 static std::string format_name_fragment(std::string_view raw)
 {
 	std::string ret;
--- a/src/naming.cc
+++ b/src/naming.cc
@@ -4,6 +4,8 @@
 // Copyright Martin Povišer <povik@cutebit.org>
 // Distributed under the terms of the ISC license, see LICENSE
 //
+#include <cctype>
+
 #include "slang/ast/symbols/VariableSymbols.h"
 #include "slang/ast/types/Type.h"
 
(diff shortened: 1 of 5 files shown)


src/slang_frontend.cc
```diff
@@ -1487,6 +1575,14 @@ struct PopulateNetlist : public TimingPatternInterpretor, public ast::ASTVisitor
 
 			// FIXME: ignores variables not driven from the sync procedure
 			VariableBits driven = sync_procedure.all_driven();
+			const ast::StatementBlockSymbol *block_symbol = ff_naming_auto_block(prologue_block, sync_body);
+			int emitted_ff_names = 0;
+			for (VariableChunk driven_chunk : driven.chunks()) {
+				const ast::Type *type = &driven_chunk.variable.get_symbol()->getType();
+				emitted_ff_names += generate_subfield_names(driven_chunk, type).size();
+			}
+			bool disambiguate_block_name = emitted_ff_names > 1;
```
povik, line 1702: I read this to mean "if we are inferring more than one wide ff instance from this procedure, prepend block name when naming the instances". Why is it conditioned this way?

povik:
> Added options:
> 
>     * `--ff-naming=legacy|signal|auto` Selects the naming scheme for inferred flip-flops. `signal` derives names only from the driven signal and subfields, while `auto` uses named always_ff blocks and after that the signal-based scheme.
> 
>     * `--ff-prefix=<prefix>` Prepends a custom prefix to inferred flip-flop names. By default this is empty for the new schemes and `$driver$` for `legacy`.
> 
>     * `--ff-suffix=<suffix>` Appends a custom suffix to non-legacy flip-flop names. The default is `_reg`, the suffix is not duplicated if the base name already ends in it.

If we need all this configurability then I would favor a single option accepting a general template throught which one could compose legacy, auto, and other schemes. Do you foresee configuring prefix/suffix, or are your needs likely covered by `--ff-naming=auto` and `signal`?

src/naming.cc
```diff
@@ -79,4 +81,108 @@ std::vector<NamedChunk> generate_subfield_names(VariableChunk chunk, const ast::
 	return ret;
 }
 
+// Format one raw hierarchy or signal name into a friendly identifier
+static std::string format_name_fragment(std::string_view raw)
```
povik: Define 'friendly'. It looks to mean avoiding certain characters

src/slang_frontend.cc
phsauter: This is supposed to use the block name iff there is one signal being assigned to (ie the block name may overwrite the signal name but only for single-signal always_ffs).  
In all other cases it should just use the block name + signal name combined.

It seemed like a sensible option to incorporate named always_ff

phsauter:
> If we need all this configurability then I would favor a single option accepting a general template throught which one could compose legacy, auto, and other schemes. Do you foresee configuring prefix/suffix, or are your needs likely covered by `--ff-naming=auto` and `signal`?

I personally would likely stick with 'auto' but after talking to some other people, some seem to prefer using strictly and only the signal names. So I think making this distinction is an option people will use.  
Then I added 'legacy' as an option to preserve the current naming scheme in case someone is relying on it.  
Since this adds `$driver$` at the beginning and I wanted to have `_reg` at the end the suffix and prefix were added and exposed in case people use other conventions (I can especially see someone using `reg_` prefix instead of the suffix.

I am not at all against using one option that allows people to pick what they want. I will think about what exactly this should expose (suggestions welcome).

src/naming.cc
phsauter: Good point, I will change it.

src/slang_frontend.cc
povik: I see, so we always use the block name, and if we are inferring multiple drivers we append signal information too

phsauter:
I can definitely improve the code coverage here, I will have another look next week.

povik:
The coverage reports are mainly informational, full coverage is not a requirement to merging

src/slang_frontend.cc
```diff
@@ -165,6 +173,87 @@ static const RTLIL::IdString module_type_id(const ast::InstanceBodySymbol &sym)
...
+		const ast::StatementBlockSymbol *block_symbol, bool disambiguate_block_name,
+		const ast::ValueSymbol &symbol, std::string_view subfield_suffix)
+{
+	auto mode = settings.ff_naming_mode();
+	std::string prefix = settings.ff_naming_prefix();
+	std::string suffix = settings.ff_naming_suffix();
+
+
+	// Avoid duplicating the configured suffix when the base name already ends in it
+	auto maybe_append_suffix = [&](std::string name) {
```
povik: Is this to avoid `_reg_reg` if designers have named their variables `foo_reg`? It's surprising behavior unless you have read the code

phsauter: Yes exactly, thats the idea.

phsauter:
I changed the interface and some parts of the implementation based on your feedback.  
There is now only one `--ff-naming` argument that still accepts `signal` and `auto` shorthands but internally everything is implemented using a templating scheme that users can directly interact with.  
Templates can use `{sep=<separator>}`, `{scope}`, `{proc}`, `{signal}`, and `{proc_auto}` to compose inferred flip-flop names.

For example the `signal` shorthand expands to `{sep=_}{scope}{signal}_reg`, while `auto` expands to `{sep=_}{scope}{proc_auto}_reg`. The `proc_auto` part uses a named procedural block when it uniquely identifies the emitted flip-flop and appends signal information when multiple flip-flops are emitted.

The main question is if this should all be documented in the help message (I think its a bit too much) or if there is a better place to put it.

PR #307, opened 2026-04-28
cmake: Add option of sourcing fmt externally

infinitymdm:
Prefer the user's installed fmt, if available.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,21 +45,30 @@ endif()
 
 include(FetchContent)
 
-# The log will show "-- Using remote fmt library", but that is misleading,
-# as no network access will be done. (Our first declaration of `fmt` wins.)
-option(FMT_INSTALL OFF)
-FetchContent_Declare(
-    fmt
-    EXCLUDE_FROM_ALL
-    SOURCE_DIR ${CMAKE_SOURCE_DIR}/third_party/fmt
-)
+find_package(fmt CONFIG QUIET)
+
+if (fmt_FOUND)
+    message(STATUS "Using system fmt")
+else()
+    message(STATUS "System fmt not found; using bundled third_party/fmt")
+
+    option(FMT_INSTALL OFF)
+    FetchContent_Declare(
+        fmt
+        EXCLUDE_FROM_ALL
+        SOURCE_DIR ${CMAKE_SOURCE_DIR}/third_party/fmt
+    )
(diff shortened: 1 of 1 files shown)


povik:
I think I wanted to package fmt to have a download-less build process. This does not introduce downloads so it looks fine. I'll wait to see if @whitequark has any comments

whitequark (commented):
The CMake changes look OK from a technical perspective. From an organizational one I would personally reject this because it introduces uncertainty in making a previously static dependency a moving one, in exchange for no benefit whatsoever, but this is not up to me.

povik:
@infinitymdm is there a reason to prefer system fmt? Is there a way to configure using system fmt by setting a cmake variable without your changes to CMakeLists.txt?

infinitymdm:
I implemented this change for distro packaging, so we don't have to ship multiple fmt libraries. There are a number of reasons this is preferred from a packaging and end-user perspective (not having to manage shipping extra licenses for bundled libs, always being sure what we're linking against, etc.).

IMO it should always be *possible* to use system libraries instead of a bundled lib anyways. It doesn't have to be the default, like I've implemented here. If the preference is to use the bundled lib by default I can modify my patch to put this behind a cmake flag.

whitequark:
There is no end-user benefit to unbundling a library like `fmt`. I am aware that packagers follow this as an absolute rule and I consider it harmful in cases like this one, where unbundling the library offers no security, functionality, or code size benefits. If you do wish to blindly follow rules your own organization decides on that offer no end-user benefit then you can just patch the package.

povik:
I don't have a strong stance on this and I'm fine with a cmake flag to support overriding where fmt comes from. It adds a degree of uncertainty when some of the user base will be using a non-vendored version of fmt but patching will do that too.

MikePopoloski:
FWIW upstream I handle this by using cmake's `FIND_PACKAGE_ARGS` feature when using FetchContent, which is basically a built-in way of using a flag to redirect a FetchContent call to a local system install via find_package instead: https://github.com/MikePopoloski/slang/blob/da377445f85bfe93345809e26d1d8150ffd6fe3b/external/CMakeLists.txt#L15

I do agree though that for small statically linked libs like fmt it's pretty silly to redirect to a system-installed shared lib, but package vendors seem to absolutely love it, so providing some option is the most user friendly.

whitequark (approved):

Issue #103, opened 2025-04-01
Alternative naming scheme for emitted modules


povik:
The names of emitted modules are formed by taking the source module name, dollar sign, and a hierarchical path to an instance. E.g. `NV_NVDLA_reset$NV_nvdla.u_partition_o.u_sync_falcon_reset` for module `NV_NVDLA_reset` instantiated at `NV_nvdla.u_partition_o.u_sync_falcon_reset`.

As we roll out feature #4 this is not a good naming scheme if the same emitted module can correspond to different instances. Instead we should have an option to form the emitted module name by attaching parameter values to the base source name, possibly further attaching `$2`, `$3` etc. to differentiate module instances equivalent in parameter values but differing e.g. in connected interfaces.

Issue #4, opened 2024-06-15
Emitting netlists with non-uniquified hierarchy


povik:
A notable shortcoming of the current frontend is that it always uniquifies the design hierarchy. Definitions of modules are copied per each instance even if the definition is identical and could be shared.

The slang AST, which we consume, already has this property. See https://github.com/MikePopoloski/slang/issues/949 for a related slang issue

povik:
Starting with 05b36c9 we now reuse modules if "instance caching" is enabled through option `--disable-instance-caching=false`. Later this will become the default

povik:
> Starting with ~[05b36c9](https://github.com/povik/yosys-slang/commit/05b36c9dc8d3885b70361f6843a0d31611657b67)~ [2a93355](https://github.com/povik/yosys-slang/commit/2a93355ef41dc20e64e4bcd4c6a09d70f18c9e76) we now reuse modules if "instance caching" is enabled through option `--disable-instance-caching=false`. Later this will become the default

At this time the feature is experimental. Known bugs:

* [ ] scopes remap doesn't match canonical body leading to error "While generating the netlist content ... wire for symbol ... is missing"

PR #309, opened 2026-04-29
Add parameter-based module uniquification

phsauter:
This is on-top of the FF-naming in #305 for my own sanity while testing things on our end.

This adds `--module-uniquify` which changes how the modules in the hierarchy are preserved when using `--keep-hierarchy` or `--best-effort-hierarchy`:
-   `--module-uniquify instance` the current behavior where every instance of the module is uniquified (meaning all module instantiations are unique)
-  ` --module-uniquify param` new behavior where every unique parametrization is uniquified (meaning some instantiations are shared if hey use the same parametrization)

The `param` mode also needs a new naming scheme for its modules which you can see from a few examples in Croc. It tries to writeout the parameters explicitly in the name but once the string gets too long it will switch to using a hash instead.
```
sync__STAGES_2__ResetValue_0
sync__STAGES_3__ResetValue_0

counter__WIDTH_4__STICKY_OVERFLOW_0
counter__WIDTH_5__STICKY_OVERFLOW_0
counter__WIDTH_16__STICKY_OVERFLOW_0
  
fifo_v3__FALL_THROUGH_0__DATA_WIDTH_32__DEPTH_2__dtype_logic_33downto0__ADDR_DEPTH_1
fifo_v3__FALL_THROUGH_0__DATA_WIDTH_8__DEPTH_16__dtype_logic_7downto0__ADDR_DEPTH_4

addr_decode__p505912c7
cdc_2phase_clearable__p86ddbee9
```

As I understand it, I am already respecting interfaces and possible port compatibility but there are cases I am not covering with any tests or have given enough thought to.

### DIFF
--- a/src/builder.cc
+++ b/src/builder.cc
@@ -24,7 +24,9 @@ static IdString id(std::string_view sv)
 #else
 static IdString id(std::string_view sv)
 {
-	return sv;
+	if (sv.empty() || sv[0] == '$' || sv[0] == '\\')
+		return sv;
+	return RTLIL::escape_id(std::string(sv));
 }
 #endif
 
--- a/src/initialization.cc
+++ b/src/initialization.cc
@@ -24,7 +24,7 @@ template <typename Func> void visit_netlist_variables(NetlistContext &netlist, F
 				}
 			},
 			[&](auto &visitor, const ast::InstanceSymbol &symbol) {
-				if (netlist.should_dissolve(symbol))
+				if (netlist.settings.should_dissolve(symbol))
 					visitor.visitDefault(symbol);
 			},
 			[&](auto &, const ast::ProceduralBlockSymbol &) {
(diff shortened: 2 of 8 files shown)


povik:
>  --module-uniquify param new behavior where every unique parametrization is uniquified (meaning some instantiations are shared if hey use the same parametrization)

There's `--disable-instance-caching=false` as an experimental option to try to reuse module definitions where possible. It's experimental because it can mishandle some edge cases related to interfaces. See https://github.com/povik/yosys-slang/issues/4.

> The param mode also needs a new naming scheme for its modules which you can see from a few examples in Croc. It tries to writeout the parameters explicitly in the name but once the string gets too long it will switch to using a hash instead.

I acknowledge we need better naming of modules when definitions are reused. Fwiw there's ticket #103. As next steps with your PR I think we need to confirm the instance caching flag does what you are after, and then only keep the module naming change.

povik:
@phsauter if you want to pursue the model naming change please open a new PR.

PR #310, opened 2026-04-30
Add Bazel build for slang.so

oharboe:
Builds the plugin against `@yosys//:hdrs` (yosys 0.62.bcr.2 onwards).
slang and fmt stay vendored as the existing submodules; the
`diagnostic_gen.py` / `syntax_gen.py` codegen runs in genrules,
mirroring what cmake does. `SLANG_VERSION_PATCH`/`HASH` are stubbed —
the cmake build derives them from `git describe` and the same hook
can be wired in here later.

Smoke-tested with bazel 8.6.0 against an in-flight yosys 0.62.bcr.2
fork (https://github.com/bazelbuild/bazel-central-registry/pull/8636);
build is cache-clean in ~40s.

Together with that yosys 0.62.bcr.2 BCR module this is enough to add
yosys-slang to the registry — the only piece still missing is a
release tarball with submodules vendored. Happy to follow up with a
GitHub release workflow that runs `git submodule update --init
--recursive` before tarring, if you'd like to be the BCR submitter.

PR #311, opened 2026-05-01
Bazel: make slang.so publicly visible

oharboe:
## Summary

`cc_shared_library` defaults to package-private visibility, so external Bazel modules can't depend on `@yosys-slang//src/yosys_plugin:slang.so` — currently the only artifact downstream actually consumes from this repo.

This blocks the natural follow-up to #310: an external module bundling the plugin into yosys's `share/` dir for plugin loading. One-line fix to set `//visibility:public` on the shared library.

## Test plan

- [x] Local `bazelisk build //src/yosys_plugin:slang.so` still succeeds.
- [x] An external project depending on `@yosys-slang//src/yosys_plugin:slang.so` now resolves (verified in The-OpenROAD-Project/OpenROAD#10237 follow-up).

### DIFF
--- a/src/yosys_plugin/BUILD.bazel
+++ b/src/yosys_plugin/BUILD.bazel
@@ -3,5 +3,6 @@ load("@rules_cc//cc:cc_shared_library.bzl", "cc_shared_library")
 cc_shared_library(
     name = "slang.so",
     shared_lib_name = "slang.so",
+    visibility = ["//visibility:public"],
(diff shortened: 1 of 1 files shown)


oharboe:
@hzeller FYI, I need this to whittle down https://github.com/The-OpenROAD-Project/OpenROAD/pull/10237

hzeller (approved):
LGTM

PR #312, opened 2026-05-05
Return const references from wire cache

gtkiku:
Copying RTLil::SigSpec turns out to potentially be very expensive for wide signals. When accessing the copy via operator[], Yosys internally calls SigSpec::unpack(), which iterates over all bits in the signal. This can potentially lead to O(n^2) behaviour when a copy is implicitly created for every bit in a signal, such as in VariableState::evaluate().

This was enough to cause a runtime difference of ~45s versus not finishing overnight for designs that happened to have a few extremely wide signals, on the order of >100KiB in multidimensional arrays.

### DIFF
--- a/src/slang_frontend.cc
+++ b/src/slang_frontend.cc
@@ -2661,7 +2661,7 @@ bool is_special_net(const ast::Symbol &symbol)
 		&& is_special_net_type(symbol.as<ast::NetSymbol>().netType);
 }
 
-RTLIL::SigSpec NetlistContext::add_wire(const ast::ValueSymbol &symbol)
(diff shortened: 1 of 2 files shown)


povik:
That's a curious finding. This seems to be a good change but with

> [...] designs that happened to have a few extremely wide signals, on the order of >100KiB in multidimensional arrays.

I'm wondering if the tool should have inferred a memory.

povik:
If you have a small reproducer at hand it would be useful to add as a regression test. I have refactoring work upcoming which will affect wire_cache and there's risk of the issue coming back.

PR #313, opened 2026-05-06
Add huge_reg test after recent performance fix

gtkiku:
Here's a minimal testcase for #312. Interestingly, seems that `some_helper` just existing stops `mem` from being inferred as regular memory, doesn't even have to be called.

Not sure how useful structures like this are, in my case I had some simulation helper functions that I had forgotten to `ifdef` out for synthesis that tripped the condition.

Changing `ADDRW = 14` changes the runtime, 14 seemed like a good compromise of not being significantly slower than the other testcases with #312 applied while still being way faster compared to before. Very wide signals are still a bit slow to elaborate, so this is among the slower tests unfortunately. Some kind of timeout would probably be useful, though.


povik:
Thanks, I have adjusted the test to have it explicitly opt out from memory inference.

> Interestingly, seems that some_helper just existing stops mem from being inferred as regular memory, doesn't even have to be called.

`InferredMemoryDetector` could be improved to ignore functions which are never called.

PR #291, opened 2026-02-13
Add support for simple concurrent assertions

gipsyh:
**This PR was fully generated by Claude Opus 4.6.**

This PR adds support for synthesizing simple concurrent assertions in SystemVerilog.

## Supported syntax

```systemverilog
assert property (@(posedge clk) expr);
assert property (@(negedge clk) expr);
assert property (@(posedge clk) disable iff (!rst_n) expr);
assume property (@(posedge clk) disable iff (!rst_n) expr);
cover property (@(posedge clk) expr);
```

Implementation

The concurrent assertion is converted to a $check cell with:
- TRG port connected to the clock signal
- TRG_POLARITY set based on posedge (1) or negedge (0)
- EN port set to the inverted disable condition (or constant 1 if no disable iff)
- FLAVOR parameter set to "assert", "assume", or "cover"

Limitations

The following SVA features are NOT supported and will still produce the "SVA unsupported" error:
- Sequence operators (##, [*], [->], etc.)
- Implication operators (|->, |=>)
- Property operators (not, and, or, until, etc.)
- Repetition in simple expressions
- Clock with iff condition (@(posedge clk iff en))

Test

Added tests/various/concurrent_assert.sv and tests/various/concurrent_assert.ys to verify the implementation.

Fixes the error for simple concurrent assertions like:
assert property (@(posedge clk) disable iff (!rst_n) count != {(W - 4) {1'b1}});

PR #317, opened 2026-05-15
Rebase SVA support into main

mlyoung101:
Pinging #77 

Very rough initial draft of trying to merge mndstrmr's work with upstream. Basically what I did was:
- Move the bulk of mndstrmr's eval code into `sva.cc`
- Plumb their `evalAssertion` routine to extract a SigSpec
- ReduceOr the SigSpec and plumb that into the existing code

This doesn't yet pass the tests, because the tests seem to hardcode an expected number of nodes in the RTLIL tree:

```
select -assert-count 9 t:$check
```

Thinking about it, I'm not sure this is the best way to verify this stuff; but also then again I don't know what the best way to verify this thing is. How can you verify the verifier? Who proves the prover? I guess we could write some circuits that we know would pass/fail and run them through SymbiYosys to check the behaviour is what we expect, and that might be sufficient. We could also compare it against Verific generated RTLIL, and/or ensure that a Verific generated circuit and a yosys-slang generated circuit will output the same result.

### DIFF
--- a/src/sva.cc
+++ b/src/sva.cc
@@ -5,6 +5,7 @@
 // Distributed under the terms of the ISC license, see LICENSE
 //
 // clang-format off
+#include "kernel/rtlil.h"
 #include "slang/ast/statements/MiscStatements.h"
 #include "slang/ast/expressions/AssertionExpr.h"
 #include "slang/ast/symbols/BlockSymbols.h"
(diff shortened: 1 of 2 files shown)


src/sva.cc
```diff
@@ -16,6 +17,263 @@
...
+		case slang::ast::AssertionExprKind::Case:
+			eval.netlist.add_diag(diag::AssertionUnsupported, expr.syntax->sourceRange().start());
+			return {};
+	}
+	log_abort(); // Unreachable
+};
+
+RTLIL::SigSpec evalAssertion(EvalContext& eval, const ast::AssertionExpr& assertion) {
+	auto paths = synthesizeAssertionExpr(eval, assertion);
+	if (paths.empty()) return false; // Ran into an error
```
mlyoung101, line 267: this doesn't seem right either we should probably add a diagnostic to the syntax tree

povik:
> This doesn't yet pass the tests, because the tests seem to hardcode an expected number of nodes in the RTLIL tree:
> 
> ```
> select -assert-count 9 t:$check
> ```
> 
> Thinking about it, I'm not sure this is the best way to verify this stuff; but also then again I don't know what the best way to verify this thing is.

I agree, ideally we would match on behavior not the number of cells. We could write explicit circuitry for the property, and then do a sequential equivalence check to confirm the hand-written circuitry and whatever the SVA property got synthesized to agree on the pass/fail verdict.

povik:
Thanks, I'll take a detailed look later.

mlyoung101:
> > This doesn't yet pass the tests, because the tests seem to hardcode an expected number of nodes in the RTLIL tree:
> > ```
> > select -assert-count 9 t:$check
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Thinking about it, I'm not sure this is the best way to verify this stuff; but also then again I don't know what the best way to verify this thing is.
> 
> I agree, ideally we would match on behavior not the number of cells. We could write explicit circuitry for the property, and then do a sequential equivalence check to confirm the hand-written circuitry and whatever the SVA property got synthesized to agree on the pass/fail verdict.

Yeah that seems like a good idea. I might also make a post and see if I can ask some of the YosysHQ devs how they verified the Verific SVA emitter.

hernloh:
@mlyoung101 @povik I'm providing a Claude authored CI fix to this PR at https://github.com/Precision-Contribution-Group/yosys-slang/pull/1 I've manually reviewed it but I'm no expert on yosys-slang, please feel free to use.

mlyoung101:
Okay, what exactly is going on here? This issue has been pinged like 10 times in the past hour or so.

This may come across as harsh, and others are free to disagree with me, and I'm not a yosys-slang maintainer, just a contributor so take my word at that. Not that I believe in the use of LLMs in general, but I especially emphasise that there are a number of places one should _absolutely never_ vibe code, and the backend of the formal proof tool (!) is one of them. This is the one section that bridges Yosys with the underlying SMT solver, that is especially important and somewhat difficult to verify. If we do not parse SVA correctly here, there is a serious risk of invalidating all our user's proofs and at worst possibly returning false SAT proofs when in reality none exist.

If you had followed the discussion in this thread, #77 and on the YosysHQ Discourse, you would know that myself and povik are both aware that the tests are failing because of the mismatched `$check` cell count. We are talking about writing a better methodology to actually check whether or not the formal circuits are equivalent, not just counting the cells. I'm also getting feedback from the actual upstream YosysHQ maintainers on this question as they've already written the Verific SVA tool. This is all currently in progress.

Again I am not a maintainer, but my current understanding (e.g. #291, #237, etc) is that they are very sceptical of the use of LLMs at all in this project; which again you would know if you engaged with the project.

You are of course free to do what you wish with your own yosys-slang fork, but I suggest that you give povik the time they need to review this PR, and me the time to complete it, and we will finish and merge it into main when appropriate.

hernloh:
Apologies to mlyoung, I did not intend to ping this issue multiple times please do not interpret this as impatience (github auto-links were causing downstream fork changes to show up here, I'll avoid auto-linking in future). I greatly appreciate your and povik's generous contribution of time to this project.

My intent was to provide additional diagnostics info in case it would be useful to either of you. As the feature branch is in the Precision-Contribution-Group, I could not send a diff easily in this repo. There are only 2 files touched, 3 lines total which a maintainer would have been able to check.

Thanks for taking time to point me to the relevant contribution guides on use of LLMs for this project.

povik:
>  We could write explicit circuitry for the property, and then do a sequential equivalence check to confirm the hand-written circuitry and whatever the SVA property got synthesized to agree on the pass/fail verdict.

I have put together a prototype at https://github.com/povik/yosys-slang/tree/sva-rebase. At first I wanted to push it here but it looks like you don't allow maintainer edits.

mlyoung101:
Thanks, I will have a look at your work.

I was confused why I couldn't enable edits by maintainers, but it looks like its a GitHub bug: https://github.com/orgs/community/discussions/5634. I keep my forks in a separate organisation so as not to clutter my main GitHub profile, and it looks like for whatever reason GitHub does not support enabling maintainers to edit forks created in orgs :/

If you want to replace this draft PR with your own draft PR and then we can collaborate over there instead, feel free.

povik:
I see. I'm not sure we won't have the opposite problem if I open a draft PR so let's keep yours. I'll start reviewing the implementation now that I drafted testing.

src/sva.cc
```diff
@@ -16,6 +17,263 @@
...
+
+	auto sig = paths[0];
+	for (size_t i = 1; i < paths.size(); i++) sig = sig || paths[i];
```
povik, line 270: @mndstrmr can I ask you to chime in and confirm my understanding: there's a bug and we should be shifting the individual elements so that their `start` matches before ORing here

povik: Tagging @defvs for any insight on his part

mndstrmr: Full disclaimer: it's been roughly a year since I looked at this code, so I'm quite rusty, but yes I think you are correct:

For example:
`a or (##10 c)`

Then there are two paths: `a` with shift 0 and `c` with shift 10. Keeping them orred without shifting will produce
```
a or c or (10 cycles haven't passed yet)
```
In the trace where `c` is never true, but `a` becomes true forever from cycle 10, then no counterexample is found despite them existing for cycles 0-9.

If we shift as you suggest, then we get
```
$past(a, 10) or c or (10 cycles haven't passed yet)
```
Here we get the counterexample we expect.

Good find! I don't really have the time to write a fix myself unfortunately, but I kinda recommend doing some testing - we tested it mostly by using it, so there's stuff we didn't cover well, e.g. the `or` SVA operator isn't used much in our FV.

povik: Thanks! This is helpful, and yes, we'll get more testing going.

mndstrmr:
I appreciate the work you both have done, apologies I've not been more involved, unfortunately I have sort of moved on from this, I'll try and help a bit where I can though.

This is minor, but given I did write the core logic myself, if this gets merged it might be nice if I was a co-author on the commit, or was otherwise listed somewhere (though I appreciate you guys have handled all of this without me!).

src/sva.cc
```diff
@@ -16,6 +17,263 @@
 
 namespace slang_frontend {
 
+// This portion was written by Louis-Emile Ploix "mndstrmr" (c) 2025; ISC licence
+// Brought into Slang head by Mel Young 2026, no additional work
```
povik, line 21: @mndstrmr see credit here. If I won't forget I'll also tag you on the relevant commit before merging

mndstrmr: Great, thanks - apologies, I must have missed that.

povik: Not at all. It's only fair you get credit

PR #319, opened 2026-05-18
Support overlapped implication in concurrent assertions

gipsyh:
## Summary

- Add support for simple overlapped implication (`a |-> b`) in concurrent SVA properties.
- Lower assert/assume implications to `!antecedent || consequent`.
- Lower cover implications to non-vacuous success, `antecedent && consequent`.
- Add regression tests for plain, `disable iff`, and cover-property cases.

Assisted-by: Codex:ChatGPT-5.5

## Testing
`./tests/run.sh` passed.

### DIFF
--- a/src/sva.cc
+++ b/src/sva.cc
@@ -16,6 +16,27 @@
 
 namespace slang_frontend {
 
+static const ast::Expression *get_simple_assertion_expr(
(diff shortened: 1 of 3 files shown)


gipsyh:
Just noticed there's a similar PR at #317, and it seems to be more fully-featured.

PR #322, opened 2026-05-24
 Avoid dissolving instances with inout ports

Kreijstal:
yosys-slang currently cannot lower inlined bidirectional port connections correctly, instead it errors during netlist conversion.

### DIFF
--- a/src/slang_frontend.cc
+++ b/src/slang_frontend.cc
@@ -3092,6 +3092,31 @@ bool NetlistContext::should_dissolve(const ast::InstanceSymbol &sym, slang::Diag
 	if (sym.isInterface())
 		return true;
 
+	if (sym.isModule()) {
+		for (auto *conn : sym.getPortConnections()) {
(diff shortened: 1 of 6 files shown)


src/slang_frontend.cc
```diff
@@ -3092,6 +3092,17 @@ bool NetlistContext::should_dissolve(const ast::InstanceSymbol &sym, slang::Diag
 	if (sym.isInterface())
 		return true;
 
+	if (sym.isModule()) {
+		for (auto *conn : sym.getPortConnections()) {
+			if (conn->port.kind == ast::SymbolKind::Port &&
+					conn->port.as<ast::PortSymbol>().direction == ast::ArgumentDirection::InOut)
+				return false;
```
povik, line 3105: Please attach a note to why_not_dissolved, if not null, to help explain why the module was chosen for preservation when one tries a hierarchical reference across the preserved boundary

povik: Nit: use port location not sym location

Issue #154, opened 2025-05-20
Make yosys-slang warnings suppressible


povik:
Adjust yosys-slang/slang so that yosys-slang warnings can be suppressed in all the same ways slang warnings can and with the same user interface.

whitequark:
I think this is problematic to implement. Slang relies on [`findDiagFromOptionName`](https://github.com/MikePopoloski/slang/blob/master/source/diagnostics/DiagnosticEngine.cpp#L25-L30) in the command line option handler, but that in turn relies on a compile time generated table of known Slang diagnostics, which obviously doesn't contain yosys-slang ones.

@MikePopoloski, do you know how the diagnostic engine could be extended to handle this use case?

PR #326, opened 2026-05-27
Allow type parameters on blackboxed modules (silent no-op on export)

RajeshT99:
## Summary

`UninstantiatedDefSymbol::handle()` walks the body of a blackboxed module to copy its parameter list onto the downstream RTLIL Cell. Today it emits `BboxTypeParameter` ("blackbox cannot have a type parameter") on any `TypeParameterSymbol` it finds, making it impossible to author a type-parametric stub for an unknown SV module.

This PR silently ignores `TypeParameterSymbol` on blackbox export. Type parameters are compile-time SystemVerilog metadata; yosys's RTLIL representation doesn't carry them anyway. The downstream Cell still receives all *value* parameters via the existing `ParameterSymbol` branch.

## Motivation

In real designs, a common pattern is to author a stub for an unknown module so the rest of the design can elaborate. CVA6 (Ariane) has this surface:

```sv
// cva6.sv:1452 — call site
cva6_hpdcache_subsystem #(
    .icache_areq_t(icache_areq_t),
    /* 19 more type parameter bindings */
    .NumPorts(NumPorts)
) i_hpdcache (...);
```

A stub matching that call site must carry the same `parameter type` declarations, but `--blackboxed-module=cva6_hpdcache_subsystem` then trips `BboxTypeParameter`. There's no workaround that lives outside yosys-slang.

## Repro

```sv
module stub
#( parameter type icache_areq_t = logic )
( /* ports */ );
  // body...
endmodule
```

```
yosys -m slang -p "read_slang --blackboxed-module=stub stub.sv"
# => stub.sv:2:20: error: blackbox cannot have a type parameter
```

After this PR: silently accepted.

## Related

Companion to a slang-side PR (MikePopoloski/slang) that narrows the `mustBeChecker` heuristic so empty port connections to unknown modules no longer flip the checker-classification path. Either PR can land independently; together they unblock CVA6 elaboration via `read_slang`.

Discovered while building the Pwrtaur power-minimization tool's corpus-validate harness against the four major permissive RISC-V cores (picoRV32, CV32E40P, Snitch, CVA6).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/slang_frontend.cc
+++ b/src/slang_frontend.cc
@@ -2129,7 +2129,16 @@ public:
 					cell->setParam(RTLIL::escape_id(std::string(symbol.name)), *converted);
 				}
 			}, [&](auto&, const ast::TypeParameterSymbol &symbol) {
-				netlist.add_diag(diag::BboxTypeParameter, symbol.location);
+				// Pwrtaur Phase 76 — Bug #2 patch.
+				// Type parameters on blackboxed modules are compile-
+				// time only; yosys's RTL representation doesn't need
+				// them. Silently ignore instead of erroring so that
+				// type-parametric stubs (e.g. cva6_hpdcache_subsystem
+				// for CVA6 corpus-validate) can act as blackboxes
+				// without yosys-slang refusing the module. The real
+				// downstream Cell carries only value parameters via
+				// the ParameterSymbol branch above.
+				(void) symbol;
 			}, [&](auto&, const ast::InstanceSymbol&) {
 				// no-op
 			}));


RajeshT99:
Companion slang-side PR filed: MikePopoloski/slang#1850 — narrows the upstream `mustBeChecker` heuristic so empty port connections no longer mis-classify a module as a checker. That PR removes the root cause that exposed this Bug #2 in the first place. The two PRs are independent and either can land first.

povik (changes requested):
I understand the use case but I'd rather ignoring the type parameter was an opt-in. If #154 were implemented we could make the "blackbox has type parameter" error suppressible. Since that is not available I'd also accept an option disabling this error in particular, e.g. `--ignore-blackbox-type-params`

PR #327, opened 2026-05-28
Implement iff conditionals in always_ff clocking block

### LINKED ISSUE #165: Support for `iff` in `always` expression
We should support:
```
module block ();
	wire clk = 0;
	wire en = 0;
	wire a = 0;
	reg y;
	always @(posedge clk iff en == 1)
		y <= a;
endmodule
```
I think this is supposed to be synthesizable (but not sure).

defvs:
See #165 

First implementation of iff conditions in always(_ff) clocking. Synthesizes as a `$dffe` (DFF with Enable pin) with the enable being comb logic right of the iff. Async, Latch, Dual-edge DFF are left unimplemented as for now; not sure of the usefulness of those as a first step.
$past works as well in blocks which have an iff gate, generating a shift reg with `$dffe`.

Equivalence tests in `dff.ys`, not sure if another file would be better, so I moved the tests in this new file since it also somewhat tests for DFF(E) generation.

Also did some tests with simple SystemVerilog code that should generate various DFFs but nothing in depth; is there a test set that could be used for further testing? First time contributing.

AI Disclosure: First pass was written with GPT-5.5 using the Codex agent. Second pass, tests, cleanup was done in depth by myself.


povik:
> Also did some tests with simple SystemVerilog code that should generate various DFFs but nothing in depth; is there a test set that could be used for further testing? First time contributing.

I don't have any suggestions beyond what is already in the CI and your dff.ys. Currently the sequential testing coverage is poor.

PR #330, opened 2026-06-10
Fix building plugin with  latest Yosys

mmicko:


whitequark:
I'm contracted by YosysHQ to deliver improvements to Yosys which would specifically enable integration with plugin CMake build systems (both ways, i.e. you should be able to build yosys-clang as a part of Yosys) and specifically include yosys-slang. I haven't looked at this PR yet but from the stuff @mmicko mentioned on Slack I think it is a fair bit different than what I had in mind

mmicko:
This is just a fix that makes current way of using `yosys-config` to actually work for windows build on oss-cad-suite, since library name changed and this is just a proper way to extract info needed for windows build.

povik:
@whitequark I'll merge this one as an interim fix and later we can have an overhaul

PR #334, opened 2026-06-10
Fast-path static element selects

ebrevdo:
When AddressingResolver has a fully defined selector, return the direct extracted window instead of constructing the dynamic raw_mux path. Keep dynamic selectors on the existing lowering path.

Add static element select coverage for endian, range, array, and struct-member cases.

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/povik/yosys-slang/pull/334).
* #339
* #338
* #337
* #336
* #335
* __->__ #334
* #333
* #332


povik:
This one we can close as already cherry-picked (2d4b0555d8eb0ef2c498e2d80a092b0f731a97d0). The PR looks to be a rebase residuum

PR #344, opened 2026-06-19
Break out the slang vendored source into its own module extension.

hzeller:
Break out the build rules into its own BUILD file, and load the sources via a MODULE.bazel extension. Put all variables w.r.t. SLANG_VERSION into one place, so it is easy to update as needed.

Also instead of third_party/ fmt library, use that from BCR.

Now, we don't need any submodules anymore in the bazel build.

This should bring yosys-slang close to be ready to be added to the bazel central registry.


povik (commented):
I'm not familiar with Bazel enough and deferring to your judgement with this change.

PR #345, opened 2026-06-19
IWYU: include necessary headers for symbols used in memory.h

hzeller:
This fails to compile in stricter environments due to missing symbols.

### DIFF
--- a/src/memory.h
+++ b/src/memory.h
@@ -5,7 +5,17 @@
 // Distributed under the terms of the ISC license, see LICENSE
 //
 #pragma once
+#include <optional>
+#include <string>
+#include <string_view>
+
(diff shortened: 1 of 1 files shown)


povik:
@hzeller is there an easy way we could replicate the stricter environment in this project's Bazel CI?

hzeller:
I noticed this when attempting to use it in a bazel central registry situation and building with `@yosys-slang//...`. I suspect what happens is that bazel looks at each header separately due to the layering check but I am not sure.

(which indicates my next path: I'd like to put this on BCR).

Maybe we should set up a `clang-tidy` CI and watch the `misc-include-cleaner` outputs ? That should cover that.

PR #346, opened 2026-06-19
Bump slang to 6175750 and update plugin code

mmicko:
This one covers API changes done recently, more as a preparation for future updates.

PR #350, opened 2026-06-22
[bug] Fix hierarchical ref in root scope segmentation fault

likeamahoney:
Hi, all!

For this case, yosys-slang segfaults while trying to resolve a hierarchical reference parent because a static variable inside a package exists outside of module scope:

```verilog
package p;
   struct { int x; } s1;
   struct { int x; } s2;
   function void f();
      int x;
   endfunction
endpackage

module m;
   import p::*;
   if (1) begin : s1
      initial begin
         s1.x = 1;
         s2.x = 1;
         f.x = 1;
      end
      int x;
   end
endmodule
```

I would prefer to use a warning instead.

### DIFF
--- a/src/diag.cc
+++ b/src/diag.cc
@@ -135,6 +135,7 @@ DiagCode ReadmemAddressOutsideOfRange(DiagSubsystem::Netlist, 1081);
 DiagCode ReadmemWordsRangeMismatch(DiagSubsystem::Netlist, 1082);
 DiagCode ReadmemBadBinaryDigit(DiagSubsystem::Netlist, 1083);
 DiagCode NoIgnoreUnknownModules(DiagSubsystem::Netlist, 1084);
+DiagCode HierarchicalRefOutsideModulesUnsupported(DiagSubsystem::Netlist, 1085);
 
 DiagGroup unsynthesizable("unsynthesizable",
 		{IffUnsupported, GenericTimingUnsyn, BothEdgesUnsupported, ExpectingIfElseAload,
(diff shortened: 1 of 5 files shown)


mmicko:
@likeamahoney This is same thing we are seeing in https://github.com/povik/yosys-slang/pull/346, right ? 
I have pointed out in comments exact commit (right after your PR is merged in slang actually) that broke this one test.  Question is, is generating error ok in this case or it should be allowed ?

likeamahoney:
> @likeamahoney This is same thing we are seeing in #346, right ? I have pointed out in comments exact commit (right after your PR is merged in slang actually) that broke this one test. Question is, is generating error ok in this case or it should be allowed ?

Hi,

I don't think this is the same issue. My fix and this problem are two completely different things.

In that commit (https://github.com/MikePopoloski/slang/commit/f06233fdd27861968268a9a0d71f6717dba0ee5a), Mike changed the type of the port connection expression to an invalid/error type. As a result, yosys-slang started crashing. Before that change, the AST was different, and the expression type was inherited from the connected port, so this situation did not occur.

Therefore, I don't believe this is related to any of my PRs.

My assumption is that this should not be fixed on the `yosys-slang` side. Instead, it should be reported and fixed in slang, similar to the issue I filed previously in slang (due to different `yosys-slang` problem) - https://github.com/MikePopoloski/slang/issues/1755. In my opinion, when `--ignore-unknown-modules` is enabled, generating an AST node with an invalid type is not the correct behavior. @MikePopoloski 

Can you create an issue in slang or should I do?

mmicko:
@likeamahoney Thank you for elaborating your response. I am still not that aware of the details of slang code, so if you could please create issue for this, I would be grateful ?

likeamahoney:
> @likeamahoney Thank you for elaborating your response. I am still not that aware of the details of slang code, so if you could please create issue for this, I would be grateful ?

Sure, I'll create an issue and post the link here.

likeamahoney:
> @likeamahoney Thank you for elaborating your response. I am still not that aware of the details of slang code, so if you could please create issue for this, I would be grateful ?

@mmicko Done in https://github.com/MikePopoloski/slang/issues/1885

mmicko:
Thank you @likeamahoney, have mentioned your other PR in `Bump slang` PR, but no actions from your side needed, it is more to point out that new parameter must be provided to tests so they can pass.

likeamahoney:
> Thank you @likeamahoney, have mentioned your other PR in `Bump slang` PR, but no actions from your side needed, it is more to point out that new parameter must be provided to tests so they can pass.

Hi! Already fixed in slang - https://github.com/MikePopoloski/slang/issues/1885

mmicko:
@likeamahoney Thanks for this,  I have already updated https://github.com/povik/yosys-slang/pull/346 to latest slang, an reverted commit so it does not remove any of existing tests.

povik:
Thanks for the fix

PR #351, opened 2026-06-22
[bug] Fix segfault exporting a blackbox with a user-defined nettype port

likeamahoney:
A port whose type is a user-defined nettype has its `DeclaredType` set up as a link, so `getTypeSyntax()` returns `nullptr`, which `export_blackbox_to_rtlil()` dereferenced unconditionally. This leads to segfault on `read_slang --empty-blackboxes` on this case:

```verilog
package NetsPkg;
  nettype real realNet;
endpackage

module rDriver
  import NetsPkg::*;
  (output realNet out);
endmodule

module top();
  rDriver r1(.out());
endmodule
```

### DIFF
--- a/tests/various/bb_nettype_port.ys
+++ b/tests/various/bb_nettype_port.ys
@@ -1,7 +1,10 @@
-# Regression: exporting an empty blackbox whose port type is a user-defined
-# nettype used to segfault. The port's DeclaredType links to the nettype's
-# declaration, so DeclaredType::getTypeSyntax() returns nullptr, which was
-# dereferenced unconditionally in export_blackbox_to_rtlil().
+# Exporting an empty blackbox whose port type is a user-defined nettype used to
+# segfault: the port's DeclaredType links to the nettype's declaration, so
+# DeclaredType::getTypeSyntax() returned nullptr and was dereferenced in
+# export_blackbox_to_rtlil(). A real-typed nettype has no faithful RTLIL bit
+# representation, so it must error out cleanly rather than crash (or silently
+# export a bogus logic net).
+test_slangdiag -expect "value of type 'real' is unsupported for conversion to bits in this context"
 read_slang --empty-blackboxes <<SV
 package NetsPkg;
   nettype real realNet;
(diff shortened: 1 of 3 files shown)


povik:
@likeamahoney I don't seem able to reproduce the crash before your fix. For me `rDriver` gets exported and `out` has a width of 64:

```
attribute \blackbox 1'1
attribute \src "<inlined>:6.8"
module \rDriver

  attribute \src "<inlined>:8.19"
  wire width 64 output 1 \out
end
```

Is the reproducer working, or does it require a different slang version?

likeamahoney:
> @likeamahoney I don't seem able to reproduce the crash before your fix. For me `rDriver` gets exported and `out` has a width of 64:
> 
> ```
> attribute \blackbox 1'1
> attribute \src "<inlined>:6.8"
> module \rDriver
> 
>   attribute \src "<inlined>:8.19"
>   wire width 64 output 1 \out
> end
> ```
> 
> Is the reproducer working, or does it require a different slang version?

Hello!

Could you please let me know which version and exactly how you reproduced the issue?
I reproduced it myself using:

```
read_slang --empty-blackboxes repro.sv
```

On the current master (commit [c9fb382](https://github.com/povik/sv-elab/commit/c9fb382af7786025308aa561ff899bfdcd5a6ac6)) and slang revision [6175750](https://github.com/MikePopoloski/slang/tree/6175750969f17289695f94be5105f6006c82eb61), I still consistently get a segmentation fault.

yosys version - `Yosys 0.63+89 (git sha1 de99d67bb)`

povik:
Thanks. I now reproduce getTypeSyntax() returning null. It does not crash for me but that's likely a platform difference.

tests/various/bb_nettype_port.ys
```diff
@@ -0,0 +1,19 @@
...
+# declaration, so DeclaredType::getTypeSyntax() returns nullptr, which was
+# dereferenced unconditionally in export_blackbox_to_rtlil().
+read_slang --empty-blackboxes <<SV
+package NetsPkg;
+  nettype real realNet;
+endpackage
+
+module rDriver
+  import NetsPkg::*;
+  (output realNet out);
```
povik, line 15: I think in this example we should error out rather than represent realNet as a 1-bit logic net.

likeamahoney: Agreed. Updated the PR. A floating (`real/shortreal`) port now errors out via `UnsupportedBitConversion` instead of being exported as a logic net. Bit-representable nettypes (e.g. `nettype logic [3:0]`) still export normally. Updated the test to cover both cases.

PR #352, opened 2026-06-23
Fix source location segfault

likeamahoney:
TBD

### DIFF
--- a/src/abort_helpers.cc
+++ b/src/abort_helpers.cc
@@ -54,7 +54,7 @@ template <typename T>
 	auto loc = source_location(obj);
 	log_assert(loc.start().buffer() == loc.end().buffer());
 
-	if (auto sm = global_compilation->getSourceManager()) {
+	if (auto sm = global_compilation->getSourceManager(); sm && loc != slang::SourceRange::NoLocation) {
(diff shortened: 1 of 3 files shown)


povik (approved):

povik:
We should get this fix in any case and I can merge it when you are ready but I see the provided test hits the assertion
```
ast_invariant(member, unpacked.bitstreamWidth == unpacked.selectableWidth);
```
which is a bug of its own and also needs fixing.

likeamahoney:
> which is a bug of its own and also needs fixing.

Sorry, I misunderstood your comment.

I agree that the assertion failure is a separate bug. Would it make sense to merge this fix and track the assertion issue separately? The test from this PR can be used as a reproducer, and I can open an issue for it.

likeamahoney:
> > which is a bug of its own and also needs fixing.
> 
> Sorry, I misunderstood your comment.
> 
> I agree that the assertion failure is a separate bug. Would it make sense to merge this fix and track the assertion issue separately? The test from this PR can be used as a reproducer, and I can open an issue for it.

Alternatively, if you would prefer, I can investigate and work on fixing the assertion failure in this PR before it is merged.

povik:
Thanks. We can merge it as-is

PR #354, opened 2026-06-25
Add test for fixed instance caching with interface ports

likeamahoney:
`--disable-instance-caching=false` crashes when a module with interface ports is instantiated against different interfaces.

slang shares a single canonical body among equivalent instances (this is correct - `InstanceCacheKey` folds interface connections into the key). The shared body's interface refs resolve against the *canonical* instance, but we built `scopes_remap` / interface port wires from the first-processed instance, so the body looked up an unremapped scope.

Set up the submodule from the canonical instance (`ref_body->parentInstance`) and wire each cell from its own instance, zipped by index. Caching off then both coincide then non-cached path unchanged.

**A simpler fix would be to just disable caching for modules with interface ports (a few lines), but that kills caching for interface-heavy designs. I took the route that keeps caching working - happy to swap for the cache-disabling version if you'd prefer the smaller change.**

### DIFF
--- a/src/slang_frontend.cc
+++ b/src/slang_frontend.cc
@@ -2360,7 +2360,20 @@ public:
 
 			RTLIL::Cell *cell = netlist.canvas->addCell(netlist.id(sym), module_type_id(*ref_body));
 			cell->set_string_attribute(ID::hdlname, netlist.hdlname(sym));
-			for (auto *conn : sym.getPortConnections()) {
+
+			// `ref_body` may be a body shared (cached) among several instances;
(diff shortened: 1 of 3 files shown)


povik:
> I took the route that keeps caching working - happy to swap for the cache-disabling version if you'd prefer the smaller change.

I haven't looked at the code change itself but I agree with making caching work with interface ports.

povik:
I ended up fixing this in 3cbbdd0063a657eda87056f5d9e72264f2759ebc but I'd like to merge your test since I don't have any.

PR #359, opened 2026-07-06
Fix compile warning

mmicko:
Just a small unused variable warning fix


povik:
Thanks. Are these an error in yosys CI? If so, it may be worth adjusting the CI here to match

mmicko:
This one was a warning, but still it would be good to make it same, will see to keep it consistent.

PR #367, opened 2026-07-15
Import blackbox parameters from RTLIL

mmicko:
Some things could be cleaner, but not yet that much into slang API. But with this in place it is possible to use FPGA primitives loaded by read_verilog.

### DIFF
--- a/src/blackboxes.cc
+++ b/src/blackboxes.cc
@@ -40,7 +40,7 @@ void import_blackboxes_from_rtlil(
 
 	BumpAllocator alloc;
 
-	auto token = [&](TokenKind kind, std::string text = "", bool space = false, bool eol = false) {
+	auto token = [&](TokenKind kind, std::string text = "", bool space = false, bool eol = false, std::optional<SVInt> value = {}) {
 		char *ptr = (char *)alloc.allocate(text.size(), 1);
 		memcpy(ptr, text.data(), text.size());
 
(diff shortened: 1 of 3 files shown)


povik (changes requested):
I have not implemented this way of parameter import because there's no way to know the type reliably, it's only a guess, and guessing it wrong can lead to incorrect elaboration.

Could we either extend `parameter_default_values` with extra metadata or add a helper on yosys side which recovers the information by looking into `AstModule`?

mmicko:
Have to check details, but slang checks just for existance of parameter by name to be able to override value. Systemverilog can specify type of parameter, but in verilog we just have parameter NAME=value, and that value will have type but need to see about width since that can be limited for parameter as well.

Anyway this is considered WIP so we can test with some real life examples.

povik:
Understood. Even if we can recover the type reliably for Verilog I think we would need to confirm the module comes from Verilog which should be possible by checking for it being an instance of `AstModule`.

mmicko:
Some things found while testing:
- can only support blackboxes for now
- if module parsed by read_verilog is used by read_slang, hierarchy will not have proper derived modules but just one with  default parameter values used
- modules with parameters will not export them into RTLIL, even if they are defined as blackboxes

with limitation to blackboxes if something is given as parameter value that is not supported it will be caught by  other tools (like place and route), even if you have types you still can have values that are not supported by FPGA.

If we need information if something is coming from read_verilog, I would rather have attribute that will add origin of netlist and we could add that in all frontends.

PR #373, opened 2026-08-07
async_pattern: match const-comparison reset conditions at widened widths

willpartcl:
## Problem

The async-load condition matcher in `interpret_async_pattern` normalizes `x == C` conditions only when the comparison is performed at 1 bit (`rst == 1'b1`), or when `C` is zero. The very common spelling

```verilog
always @(posedge clk or posedge rst)
    if (rst == 1) q <= 1'b0;   // unsized literal: comparison at 32 bits
    else q <= d;
```

compares at 32 bits (IEEE 1800-2017 §11.6.1), so the equality is never stripped and elaboration fails with

```
error: condition cannot be matched to any signal from the event list
```

on spec-legal RTL that commercial tools accept. Hit in the wild by OpenPiton (`piton/design/chipset/noc_sd_bridge/rtl/byte_en_reg.v`), and `!=` conditions (`if (rst != 0)`, negedge `if (rst_n != 1)`) are not handled at all.

## Fix

Generalize the two constant-comparison rules soundly: for `x == C` / `x != C` with a known constant `C` on either side, look through the widening integral conversions on the signal side and reduce the comparison to `x` or `!x` exactly when `C` is the image of exactly one of `1'b1` / `1'b0` under those conversions (zero- or sign-extension per each conversion step's source type).

- Accepts `rst == 1`, `rst != 0`, `1 == rst`, `rst_n == 0`, `rst_n != 1`, and the signed sign-extension image (`s == -1` for 1-bit signed `s`).
- Still rejects constant conditions like `rst == 2` (image of neither value) and the sign-extension trap `s == 1` for 1-bit signed `s` (constant false), so the match stays semantically implied by the edge polarity.
- Subsumes the previous 1-bit-true and zero-constant special cases unchanged.

## Validation

Developed and fully validated at f3804697 (the commit partcleda/boson pins; RTLIL for `if (rst == 1)` is byte-identical modulo `src` attributes to the `if (rst)` spelling, full test suite green there, 52/52). This branch is the clean cherry-pick onto current master (context applied without conflicts); please let CI confirm on master's vendored slang. New test: `tests/various/aload_cond_const_cmp.ys` covering all accepted spellings plus the two must-still-reject cases.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/async_pattern.cc
+++ b/src/async_pattern.cc
@@ -199,26 +199,73 @@ void TimingPatternInterpretor::interpret_async_pattern(const ast::ProceduralBloc
 				did_something = true;
 			}
 
+			// Strip an (in)equality comparison against a constant, i.e. rewrite
+			// `x == C` / `x != C` into `x` or `!x` when that is sound. The
+			// signal operand may have been widened for the comparison (e.g. in
+			// `rst == 1` the comparison is performed at 32 bits per §11.6.1),
+			// so look through widening conversions on the signal side and
+			// extend 1'b0 / 1'b1 the same way those conversions would: the
+			// comparison reduces to `x` (or `!x`) iff C is the image of
+			// exactly one of the two values.
 			if (condition->kind == ast::ExpressionKind::BinaryOp &&
-					condition->as<ast::BinaryExpression>().op == ast::BinaryOperator::Equality &&
-					condition->as<ast::BinaryExpression>().right().getConstant() &&
-					condition->as<ast::BinaryExpression>().right().type->getBitWidth() == 1 &&
-					!condition->as<ast::BinaryExpression>().right().getConstant()->hasUnknown() &&
-					condition->as<ast::BinaryExpression>().right().getConstant()->isTrue()) {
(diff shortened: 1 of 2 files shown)


povik:
I agree it would be good to extend the detection but I can't look at this change as it's AI generated (see policy documented in README).

willpartcl:
I mean if its a good change and passes CI + its only 64 lines of code its a bit of a ridiculous rule no?

povik:
I don't know it's a good change until I review it, including the comments. 64 lines of code is a lot in the context of a synthesis frontend developed carefully in order not to introduce correctness bugs.

The rule may seem ridiculous but there's real asymmetry to accepting LLM generated changes - they are easy to generate and take effort (like any change) to review. I accept changes I generate myself or which come from sources attached to funding. I have found it's not sustainable to review such changes coming from the public.

povik:
@willpartcl e48b9bbd34e1699a2ea13f897f1600bcdc47ea59 extends the detection and makes your example go through

PR #382, opened 2026-09-14
Add partial support for stable,changed,rose,fell system functions

obetschi-antmicro:
Add support for `$stable` `$changed` `$rose` and `$fell` by following the same logic used for `$past`.
No support for the `clocking_event` argument (similar to `$past`).

Related to [issue 84](https://github.com/povik/sv-elab/issues/84) but not sufficient yet to solve it.


povik (commented):
Code looks good but we should have tests. Please take a look at `dff.ys` for inspiration. Continuing the pattern there one could for example do:

```
design -reset
read_slang <<EOF
module dff_rose01_gate(input logic clk, input logic d, output logic q);
	always_ff @(posedge clk) begin
		q <= $rose(d);
	end
endmodule
EOF

read_slang <<EOF
module dff_rose01_gold(input logic clk, input logic d, output logic q);
	always_ff @(posedge clk) begin
		q <= d && !$past(d);
	end
endmodule
EOF

async2sync
miter -equiv -make_assert -flatten dff_rose01_gold dff_rose01_gate dff_rose01_miter
hierarchy -top dff_rose01_miter
sat -seq 20 -set-init-zero -prove-asserts -verify dff_rose01_miter
```

